PR #3, opened 2023-06-02
NON-10 (part of): Skeleton of API endpoint and test + OAuth2 basic config (for local development)

xoen:
Mostly copied from Incentives API but required a lot of trial and error to get it working.

**NOTE**: I tried to copy as little as possible to get this simple resource test running, there is likely more Spring Boot setup required ahead.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,11 +4,33 @@ plugins {
 }
 
 configurations {
+  implementation { exclude(module = "spring-boot-starter-web") }
+  implementation { exclude(module = "spring-boot-starter-tomcat") }
   testImplementation { exclude(group = "org.junit.vintage") }
 }
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
+  implementation("org.springframework.boot:spring-boot-starter-security")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+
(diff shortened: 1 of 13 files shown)


build.gradle.kts
```diff
@@ -4,11 +4,33 @@ plugins {
...
   testImplementation { exclude(group = "org.junit.vintage") }
 }
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
+  implementation("org.springframework.boot:spring-boot-starter-security")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core")
```
neilmendum, line 18: Are we thinking to use coroutines and reactive for HTTP and DB calls?

xoen: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/3/commits/96180ba5dd4d3f67f1917ab2ce00642e9299cbca

Thoughts?

neilmendum: All sounds good to me 😄

neilmendum: Ah looks like we sent messages at the exact same time - https://github.com/ministryofjustice/hmpps-non-associations-api/pull/3#discussion_r1217735851 looks good

neilmendum (approved):
LGTM

PR #11, opened 2023-06-13
NON-10: 2nd endpoint, better docs, catch 404s, etc...

xoen:
- added `GET /legacy/api/offenders/{prisonerNumber}/non-association-details`
- deprecated `GET /legacy/api/bookings/{bookingId}/non-association-details` as we should keep things simple and always get non-associations by prisoner number (rather than by bookingId which can change when someone leaves/return to prison)
- better Swagger docs for the endpoints
- read non-associations api OAuth client creds from kubernetes `Secret`
- handle 404 Not Found from Prison API by returning the same status code rather than responding with a 500 Internal Server Error


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.justice.digital.hmpps.hmppsnonassociations.config
+package uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config
```
xoen, line 1: This was a sneaky one, because of a different package name (even tho it was in the correct directory) the exception handling was not working.

ashleygyngell: Nice spot

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
```diff
@@ -10,32 +10,55 @@ import org.springframework.web.bind.annotation.GetMapping
...
       ApiResponse(
-        responseCode = "400",
-        description = "Incorrect data specified to return list of non-associations",
+        responseCode = "401",
+        description = "Unauthorized to access this endpoint",
         content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
       ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Booking ID not found",
```
xoen, line 39: Hang on, this should be Prisoner Number. I'll sort.

xoen: 3148041cedf3945a5f129ca575bfa5c320d424cf

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
```diff
@@ -10,32 +10,55 @@ import org.springframework.web.bind.annotation.GetMapping
...
 import org.springframework.web.bind.annotation.RestController
-import uk.gov.justice.digital.hmpps.hmppsnonassociations.config.ErrorResponse
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.ErrorResponse
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.prisonapi.NonAssociationDetails
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.service.NonAssociationsService
 
-// TODO: Check endpoint format/document it
-
 @RestController
 @RequestMapping("/legacy/api", produces = [MediaType.APPLICATION_JSON_VALUE])
```
Mjwillis, line 18: Don't think it should be called legacy.  We'll want all service to switch to this and then underneath the hood switch the data source (with feature switch) - we don't want services to have to change twice.

Suggest change to `/prisoners/{prisonerNumber}/non-association-detatils`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
```diff
@@ -10,32 +10,55 @@ import org.springframework.web.bind.annotation.GetMapping
...
+  )
+  suspend fun getDetailsByPrisonerNumber(
+    @Schema(description = "The offender prisoner number", example = "A1234BC", required = true)
+    @PathVariable
+    prisonerNumber: String,
+  ): NonAssociationDetails {
+    return nonAssociationsService.getDetails(prisonerNumber)
+  }
+
+  @GetMapping("/bookings/{bookingId}/non-association-details")
```
Mjwillis, line 52: Think we should NOT implement this because WE ARE NOT GOING TO RECORD `bookingId` in our service

Mjwillis (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
xoen: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/11/commits/a4e2a49ff205cde1ae84ef2d5a45f7851335d5fe

ushkarev (approved):

PR #13, opened 2023-06-14
NON-10: Optional assigned living unit fields

xoen:
The assigned living unit fields could be null and in general a non-association could be with someone who left prison (but may return, and that's why it's imporant to still retain the non-association)

I've also update the Swagger documentation for the fields accordingly.

**NOTE**: While I was getting an exception when Spring Boot/Jackson for the missing mandatory `String` field, it didn't complain about the missing `Long` field and implicitly converted it to `0`.

We should distinguish between `0`s and `null`s are these are different concept, one signifify the absence of information (e.g. there is no assigned unit) while the other imply knowledge (e.g. the assigned unit is XYZ). We don't want `0` to be possibly interpreted as an ID for a real living unit as that would be wrong.

That's why the `Long?` optional fields have a default value of `null` now.

PS: I've updated the CircleCI config to allow deployment to `preprod`/`prod` now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,38 +63,38 @@ workflows:
             - build_docker
             - helm_lint
           helm_timeout: 5m
-#      - request-preprod-approval:
-#          type: approval
-#          requires:
-#            - deploy_dev
-#      - hmpps/deploy_env:
-#          name: deploy_preprod
-#          env: "preprod"
-#          jira_update: true
-#          jira_env_type: staging
-#          context:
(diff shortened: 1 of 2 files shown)


ushkarev (dismissed):

xoen:
Sorry @ushkarev - I've pushed another little commit after your approval to enable `preprod`/`prod` deployment via CircleCI, little point making another PR just for that IMHO.

ushkarev (approved):

PR #18, opened 2023-06-19
NON-10 Fix up the app to work non-reactively

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/WebClientConfiguration.kt
```diff
@@ -1,73 +1,85 @@
...
 
-  @Bean
-  fun prisonHealthWebClient(): WebClient {
-    return WebClient.builder()
-      .baseUrl(prisonRootUri)
-      .build()
+  private fun createForwardAuthWebClient(builder: WebClient.Builder, url: @URL String): WebClient = builder.baseUrl(url)
+    .filter(addAuthHeaderFilterFunction())
+    .clientConnector(
+      ReactorClientHttpConnector(HttpClient.create()),
```
petergphillips, line 63: normally people add a `clientConnector` to set ` responseTimeout` on the `HttpClient`, otherwise I think the default is the same as what you've provided here anyway

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/helper/JwtAuthHelper.kt
```diff
@@ -4,15 +4,14 @@ import io.jsonwebtoken.Jwts
...
+import org.springframework.security.oauth2.jwt.JwtDecoder
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder
 import org.springframework.stereotype.Component
 import java.security.KeyPair
 import java.security.KeyPairGenerator
 import java.security.interfaces.RSAPublicKey
 import java.time.Duration
-import java.util.Date
-import java.util.UUID
+import java.util.*
```
ushkarev, line 14: i thought ktlint doesn't like star imports, but this seems to pass! i looked in incentives and couldn't see what was calling ktlint that perhaps _isn't_ calling it in this ci pipeline

Mjwillis: Hmmm, `gradle check` does call it

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/AuditorAwareImpl.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config
+
+import org.springframework.context.annotation.Configuration
+import org.springframework.data.domain.AuditorAware
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing
+import org.springframework.stereotype.Service
+import java.util.Optional
+
+@Configuration
+@EnableJpaAuditing(auditorAwareRef = "auditorAware")
```
Mjwillis, line 10: FYI this now works

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
```diff
@@ -5,22 +5,29 @@ import io.swagger.v3.oas.annotations.media.Content
...
 
 @RestController
+@Validated
 @RequestMapping("/legacy/api", produces = [MediaType.APPLICATION_JSON_VALUE])
 @Tag(name = "Non-Associations", description = "Retrieve non association details from prison-api")
 class PrisonApiResource(
   val nonAssociationsService: NonAssociationsService,
 ) {
   @GetMapping("/offenders/{prisonerNumber}/non-association-details")
+  @PreAuthorize("hasRole('ROLE_NON_ASSOCIATIONS')")
```
ushkarev, line 29: is this a role that's already set it up in hmpps-auth somewhere? the `hmpps-non-associations` client doesn't have it in dev (that's the only place i can check)

Mjwillis: I'll set it up

build.gradle.kts
```diff
@@ -1,41 +1,44 @@
...
 }
 
 configurations {
-  implementation { exclude(module = "spring-boot-starter-web") }
-  implementation { exclude(module = "spring-boot-starter-tomcat") }
   testImplementation { exclude(group = "org.junit.vintage") }
 }
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
```
xoen, line 13: Probably a stupid question but is `webflux` still needed for a non-reactive project? I thought webflux was the reactive routing bit? 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/PrisonApiService.kt
```diff
@@ -15,11 +15,12 @@ class PrisonApiService(
...
   }
 
-  suspend fun getNonAssociationDetails(prisonerNumber: String, useClientCredentials: Boolean = false): NonAssociationDetails {
+  fun getNonAssociationDetails(prisonerNumber: String, useClientCredentials: Boolean = false): NonAssociationDetails {
     return getClient(useClientCredentials)
       .get()
       .uri("/api/offenders/$prisonerNumber/non-association-details")
       .retrieve()
-      .awaitBody()
+      .bodyToMono(object : ParameterizedTypeReference<NonAssociationDetails>() {})
```
xoen, line 23: What is this doing 😬?

Is this equivalent to `.bodyToMono(NonAssociationDetails::class.java)`? That seems to compile and work 🤔 (and it looks slightly less weird)

src/main/resources/application.yml
```diff
@@ -47,12 +52,6 @@ spring:
       connection-timeout: 1000
       validation-timeout: 500
 
-  flyway:
```
xoen, line 50: So `flyway` is not used anymore or it's clever enough to get the settings from the `datasource` properties?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
petergphillips: Prison API endpoints often work off the relationship between a user and the prisoner rather than specific roles.  For that endpoint see https://github.com/ministryofjustice/prison-api/blob/fb3e3e69317a88512224778b28ff248f46bf1e35/src/main/java/uk/gov/justice/hmpps/prison/service/OffenderNonAssociationsService.java#L47.  The Prison API swagger docs go into more detail.

PR #21, opened 2023-06-20
NON-10 Run in transaction to check persistence - Not last modified ti…

Mjwillis:
…me does not work in tests.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/ReasonTypeRepositoryTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/ReasonTypeRepositoryTest.kt
@@ -7,17 +7,19 @@ import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabas
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest
 import org.springframework.context.annotation.Import
 import org.springframework.security.test.context.support.WithMockUser
-import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.context.transaction.TestTransaction
+import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.AuditorAwareImpl
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.AuthenticationFacade
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.helper.TestBase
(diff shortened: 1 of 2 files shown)


src/test/resources/application-test.yml
```diff
@@ -17,6 +17,9 @@ spring:
     username: non_associations
     password: non_associations
 
+  main:
+    allow-bean-definition-overriding: true
```
ushkarev, line 21: i tried to read about what this is doing and i'm still unsure how this helps with transactions and/or annotations on the entities. is this here just to match other projects?

ushkarev (approved):

src/test/resources/application-test.yml
Mjwillis: It might not be needed now.  I take out in next PR

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/ReasonTypeRepositoryTest.kt
```diff
@@ -31,19 +34,22 @@ class ReasonTypeRepositoryTest : TestBase() {
...
-    var whenCreated = reason.whenCreated
-
+    val reason = repository.findById("VIC").orElseThrow { Exception("ReasonType not found") }
+    val whenCreated = reason.whenCreated
     repository.save(reason.copy(description = "UPDATED"))
 
-    var reasonUpdated = repository.findById("VIC").orElseThrow { Exception("ReasonType not found") }
+    TestTransaction.flagForCommit()
+    TestTransaction.end()
+    TestTransaction.start()
```
xoen, line 50: How does this work? Does it mean that by default a `@Test` method is wrapped in a transaction or something?

And if that's the case how does reading after a save in a test work? 🤔  Sorry too many stupid Spring Boot questions in these PRs

src/test/resources/application-test.yml
ushkarev: oh i'm happy to leave it if it's necessary for other things or helps align with other projects

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/ReasonTypeRepositoryTest.kt
Mjwillis: Yes `@JpaDataTest` sets up a transaction and allows tx to be commited.  JPA doesn't actually issue SQL unless the transaction is committed.  So without this bit you aren't actually issuing an INSERT/UPDATE

PR #22, opened 2023-06-21
NON-10: Skeleton of new non-associations endpoint

xoen:
- `GET /prisoner/{prisonerNumber}/non-associations`
- this is just a placeholder/starting point and very much subject to change (and this is made explicit in the Swagger docs)
- endpoint uses `ROLE_NON_ASSOCIATIONS` as discussed although authorisation may need further discussion
- the response format is just a stub at this stage (as design is still in flux and this may affect data model) - but the main point is that this format/dto class is different and separate from the one used by the legacy/Prison API endpoint
  - This is mainly to allow us design freedom, however those two formats will have to have some level of compatibility for synchronisation purposes obviously
- handling of `AccessDenied` exceptions so that the client gets an appropriate `403 Forbidden` rather than a `500 Internal Server Error`

Also, replaced `.bodyToMono(object : ParameterizedTypeReference<NonAssociationDetails>() {})` with `.bodyToMono(LegacyNonAssociationDetails::class.java)` as it seems to work the same (in this context at least) but reads better.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/ReasonType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/ReasonType.kt
@@ -19,6 +19,4 @@ data class ReasonType(
   val whenCreated: LocalDateTime = LocalDateTime.now(),
   @LastModifiedDate
   val whenUpdated: LocalDateTime = LocalDateTime.now(),
-) {
-  constructor() : this("", "")
-}
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/ReasonType.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/ReasonType.kt
@@ -18,5 +18,5 @@ data class ReasonType(
   @CreatedDate
   val whenCreated: LocalDateTime = LocalDateTime.now(),
   @LastModifiedDate
-  val whenUpdated: LocalDateTime = LocalDateTime.now(),
+  var whenUpdated: LocalDateTime = LocalDateTime.now(),
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
@@ -48,7 +48,7 @@ class PrisonApiResource(
       ),
     ],
   )
-  fun getDetailsByPrisonerNumber(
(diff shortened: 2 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/ReasonType.kt
```diff
@@ -18,7 +18,5 @@ data class ReasonType(
   @CreatedDate
   val whenCreated: LocalDateTime = LocalDateTime.now(),
   @LastModifiedDate
-  val whenUpdated: LocalDateTime = LocalDateTime.now(),
-) {
-  constructor() : this("", "")
```
ushkarev, line 23: yay

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
```diff
@@ -15,13 +15,13 @@ import org.springframework.web.bind.annotation.RequestMapping
...
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.ErrorResponse
-import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.prisonapi.NonAssociationDetails
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.prisonapi.LegacyNonAssociationDetails
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.service.NonAssociationsService
 
 @RestController
 @Validated
 @RequestMapping("/legacy/api", produces = [MediaType.APPLICATION_JSON_VALUE])
-@Tag(name = "Non-Associations", description = "Retrieve non association details from prison-api")
+@Tag(name = "Legacy non-associations-details", description = "Retrieve non association details from Prison API")
```
ushkarev, line 24: should we remove `@PreAuthorize("hasRole('ROLE_NON_ASSOCIATIONS')")` from this, the old api because it's not how prison-api did permissions?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/PrisonApiResource.kt
xoen: Yeah good call, I'll do 👍

xoen: Removed that role requirement for the legacy endpoint: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/22/commits/07ea92464af5ad98b6b9de39bd737cad4e473cb9

ushkarev (dismissed):

ushkarev (approved):

PR #27, opened 2023-07-03
NON-77: Non-association starter model

xoen:
Basic DB schema for a non-association between two prisoners.

We may tweak the data model as we know more about the requirements but the basics are there:
- details about the 2 prisoners to non-associate (prisoner numbers and reason, e.g. "Perpetrator", etc...)
- why they're should be kept apart, e.g. comment and optional incident report number
- type of restriction, e.g. "don't locate in the same cell"
- who authorised this non-association
- whether the non-association is closed and why/by who/when (by default non-associations are open)

**NOTE**: The `reason_type` and `restriction_type` DB tables were dropped and replaced by simple enums now.


Mjwillis (approved):
LGTM

PR #28, opened 2023-07-07
NON-90 Publish domain events when a non-association is created/updated

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/AuditService.kt
```diff
@@ -0,0 +1,70 @@
...
+  private val auditQueueUrl by lazy { auditQueue.queueUrl }
+
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun sendMessage(auditType: AuditType, id: String, details: Any, username: String? = null) {
+    val auditEvent = AuditEvent(
+      what = auditType.name,
+      who = username ?: authenticationFacade.currentUsername ?: "non-associations-api",
```
xoen, line 35: There is a `SYSTEM_USERNAME` constant which is probably better to use for consistency.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/AuditService.kt
```diff
@@ -0,0 +1,70 @@
...
+      service = serviceName,
+      details = objectMapper.writeValueAsString(details),
```
xoen, line 37: I see there is a `toJson()` method added to `Any`, does it mean you could use it on `details` too here?

e.g. `details = details.toJson()`. It would probably read more nicely ☺️

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SnsService.kt
```diff
@@ -0,0 +1,98 @@
...
+    additionalInformation,
+    "1.0",
+    occurredAt.toOffsetDateFormat(),
+    description,
+  )
+}
+
+enum class NonAssociationDomainEventType(val value: String, val description: String, val auditType: AuditType) {
+  NON_ASSOCIATION_CREATED("non-association.created", "A non-association has been created: ", AuditType.NON_ASSOCIATION_CREATED),
+  NON_ASSOCIATION_UPSERT("non-association.amended", "A non-association has been amended: ", AuditType.NON_ASSOCIATION_UPDATED),
```
xoen, line 94: namespace should be plural, e.g. `non-associations.*`

I'd also call the event `NON_ASSOCIATION_UPDATED("non-associations.updated", ...`

Sorry, don't mean to be too picky, we could always rename later on but it's a bit tricker with events stuff.

xoen (changes requested):
This looks great, thanks!

Only thing I'd tweak before merging - [see comment](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/28/files#r1257993059) - is the name of the domain events published to be plural (`non-associations.*`) and to call the event just "update".

PS: I didn't just approve as I know you would have merged it straight away if I did 😝

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/AuditService.kt
Mjwillis: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/AuditService.kt
Mjwillis: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SnsService.kt
Mjwillis: I've updated to plural but I think the convention is to used created and amended now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -10,6 +10,7 @@ import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociation a
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.NonAssociation as NonAssociationJPA
 
 @Service
+@Transactional
```
xoen, line 13: 🤔 

I know this service doesn't do much but my idea was to not have a blanket `@Transactional` for all methods, and add it only for methods that needed it.

With that in mind, a method like `createNonAssociation()` would not be trasactional, it would use another method to persist the record and then publish a domain event (which we know would be problematic if done within a transaction as if for whatever reason there is a problem and the transaction is rolled back it whoever process it may get a 404)

Does this make sense? I would prefer to only have the `persistNonAssociation()` method to be transactional to avoid these kind of problems (and this is simpler than using more complicated things like using a separate class for messaging)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SnsService.kt
xoen: Really? That's odd, alright I guess 😅

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
xoen: Just to be clear, the actual sending of domain events would be in its own class, but invoking these would happen in the `NonAssociationsService` rather than having to wrap `NonAssociationsService` inside some other class just to avoid the transactional problems - if that make sense.

Happy to have a chat if that is not clear.

xoen: That's why there was that [`.also {}` placeholder](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/28/files#diff-46b4197c69d9454a21c6abbd82fb1663e6bc1f26f53cfa8695bbd77a8b4393f9L23-L25) in the service.

the `createNonAssociation()` method (non-transactional) would use `persistNonAssociation()` (transactional) _**and then**_ publish the domain event to notify other services (outside the transaction, once this is committed)

Mjwillis: Yes makes sense but I've wrapped the service call with the event /audit in the controller resource.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -20,16 +21,13 @@ class NonAssociationsService(
...
-      // TODO: Publish domain event (outside transaction)
-    }.toDto()
+    return persistNonAssociation(nonAssociationJpa).toDto()
   }
 
   fun getDetails(prisonerNumber: String): LegacyNonAssociationDetails {
     return prisonApiService.getNonAssociationDetails(prisonerNumber)
   }
 
-  @Transactional
```
Mjwillis, line 32: you can't put `@Transactional` on a private method

xoen (approved):

PR #29, opened 2023-07-07
NON-86: `POST /non-associations` endpoint

xoen:
- ability to create a non-association (open). required `ROLE_NON_ASSOCIATIONS` role with `write` scope
- better handling of some validation exceptions
- non-association `comment` is mandatory in JPA entity (was already non-nullable in DB)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
@@ -14,6 +14,7 @@ import org.springframework.data.jpa.domain.support.AuditingEntityListener
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociationReason
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociationRestrictionType
 import java.time.LocalDateTime
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociation as NonAssociationDTO
 
 @Entity
 @EntityListeners(AuditingEntityListener::class)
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
```diff
@@ -57,4 +58,20 @@ class NonAssociation(
...
+      firstPrisonerNumber = firstPrisonerNumber,
+      firstPrisonerReason = firstPrisonerReason,
+      secondPrisonerNumber = secondPrisonerNumber,
+      secondPrisonerReason = secondPrisonerReason,
+      restrictionType = restrictionType,
+      comment = comment,
+      // TODO: Do we need to do anything special with this?
+      //       This field being optional in NOMIS/Prison API
+      //       It may be one of the things we make mandatory after migration?
+      authorisedBy = authorisedBy ?: "",
```
Mjwillis, line 74: This isn't a field at the moment in the design I believe.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
```diff
@@ -57,4 +64,39 @@ class NonAssociationsResource() {
...
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Missing required privileges. Requires the NON_ASSOCIATIONS role with write scope",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun createNonAssociation(
+    @RequestBody
```
Mjwillis, line 97: put `@validated`

Mjwillis (approved):
This is a good starting point!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
xoen: I've named this `authorisedBy` as it's called in Prison API/NOMIS.
It seems to be free text and optional in Prison API but for records created via Non-associations API this field would be populated with the username of the user who created it.

PR #34, opened 2023-07-11
NON-90 Handle Prisoner Number merges from NOMIS

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
```diff
@@ -18,7 +18,7 @@ import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociation a
 
 @Entity
 @EntityListeners(AuditingEntityListener::class)
-class NonAssociation(
+data class NonAssociation(
```
Mjwillis, line 21: Without this couldn't clone the object!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -36,4 +42,18 @@ class NonAssociationsService(
   private fun persistNonAssociation(nonAssociation: NonAssociationJPA): NonAssociationJPA {
     return nonAssociationsRepository.save(nonAssociation)
   }
+
+  fun mergePrisonerNumbers(oldPrisonerNumber: String, newPrisonerNumber: String): List<NonAssociationJPA> {
```
ushkarev, line 46: this doesn't persist the changes does it?

ushkarev: actually, even when it does, there could be a separate problem where the newly created copy matches a non-association pair in the DB. imagine:
• A has a non-association to B and another to C
• B is merged into C
• now we have two non-associations between A and C

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SnsService.kt
```diff
@@ -62,10 +62,9 @@ class SnsService(hmppsQueueService: HmppsQueueService, private val objectMapper:
 }
 
 data class AdditionalInformation(
```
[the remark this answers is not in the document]
Mjwillis: yes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
Mjwillis: Yes I did think of this, its needs more work to handle that!

Mjwillis: Also the persistant bit was missing which I added this morning- sorry remember after the commit yesterday!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
xoen: I haven't looked at the PR yet but the choice of _**not**_ using data classes _for entities_ was deliberate.

IIRC docs mention to not use data classes for JPA entities, it was also mentioned to avoid doing so [in this article](https://jpa-buddy.com/blog/best-practices-and-common-pitfalls/) and when I was playing with some entities some time ago I was noticing some strange behaviour (but it was some time ago and I don't remember exactly sorry)

Are we sure this is right to do and everything will continue to work as expected (all the usual JPA machinery) even tho is a bit less convenient? (I appreciate you can't copy a entity but it can be updated if necessary, and if a field should be modified we can change to a `var` field, if it makes sense)

ushkarev: oh yeah i forgot about this. the spring docs do definitely say you should not use immutable classes (i.e. data classes) with JPA because the proxy wizardry silently fails sometimes.

for the purpose of this PR, you really are mutating an existing record so there's no requirement to make a copy of it. you can just change it in place.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -36,4 +42,18 @@ class NonAssociationsService(
...
+    log.info("Replacing prisoner number $oldPrisonerNumber to $newPrisonerNumber")
+
+    val nonAssociationList1 = nonAssociationsRepository.findAllByFirstPrisonerNumber(oldPrisonerNumber).map {
+      it.copy(firstPrisonerNumber = newPrisonerNumber)
```
xoen, line 50: if `firstPrisonerNumber`/`secondPrisonerNumber` are changed to be `var` (as they should as they need to be updated now) you shouldn't need to `copy()`, you could just update the instance with `it.firstPrisonerNumber = newPrisonerNumber`

(so you don't need to make the entity a data class which could lead to unexpected issues with JPA)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
xoen: yeah agreed, change the `firstPrisonerNumber`/`secondPrisonerNumber` fields to be mutable (`var` instead of `val`) should do and it's nice because they could change (as in this case of a merge event which update records' prisoner numbers)

xoen (dismissed):
This looks good ☺️, only thing that I'd change is to keep the `NonAssociation` JPA entity as a normal class as it was by making `firstPrisonerNumber`/`secondPrisonerNumber` mutable fields with `var` (see comments)

ushkarev:
or perhaps for more consistency, all fields should be mutable?

xoen:
> or perhaps for more consistency, all fields should be mutable? we're bound to need to mutate other parts in future
My preference would be to change only the fields that need to change to mutable. Although in the case of a non-association possibly most of them would need to change (but not all).

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SnsService.kt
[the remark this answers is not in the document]
Mjwillis: I'l like to sort this in next PR

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
```diff
@@ -74,4 +77,20 @@ data class NonAssociation(
...
+
+    if (firstPrisonerNumber != other.firstPrisonerNumber) return false
+    return secondPrisonerNumber == other.secondPrisonerNumber
+  }
+
+  override fun hashCode(): Int {
+    var result = firstPrisonerNumber.hashCode()
+    result = 31 * result + secondPrisonerNumber.hashCode()
+    return result
+  }
```
xoen, line 95: Just out of curiosity, what are these methods for? Are they needed for some of the functionality that was added or more like something that could be useful in the future?

src/main/resources/db/migration/V1_9__add_unique_constraint.sql
```diff
@@ -0,0 +1,3 @@
+-- Increase length of `authorised_by` column to 60 chars as in NOMIS
```
ushkarev, line 1: not really the right comment!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationMergeServiceIntTest.kt
```diff
@@ -0,0 +1,106 @@
...
+        firstPrisonerNumber = "B1234AA",
+        secondPrisonerNumber = "A1234AA",
+        createTime = createTime,
+      ),
+    )
+
+    val mergedAssociations = service.mergePrisonerNumbers("A1234AA", "B1234AA")
+
+    Assertions.assertThat(mergedAssociations.toSet()).isEqualTo(
+      setOf(
```
xoen, line 92: I imagine the `equals()` and `hashCode()` methods were added in order for this to work. I guess it's useful for testing purposes when we don't care for the specific records being updated/deleted 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
```diff
@@ -0,0 +1,79 @@
...
+      .plus(nonAssociationsRepository.findAllBySecondPrisonerNumber(oldPrisonerNumber)).forEach { nonAssociation ->
+
+        log.debug("Looking at record ${nonAssociation.id}")
+        if (nonAssociation.firstPrisonerNumber == oldPrisonerNumber) {
+          nonAssociationsRepository.findByFirstPrisonerNumberAndSecondPrisonerNumber(
+            firstPrisonerNumber = newPrisonerNumber,
+            secondPrisonerNumber = nonAssociation.secondPrisonerNumber,
+          ).let {
+            if (it != null) {
+              nonAssociationsRepository.delete(nonAssociation)
```
xoen, line 36: Do we need to publish a domain event when a non-association is deleted?

ushkarev: omg. i just realised that syncing this will be a nightmare. because a merge comes from nomis, it'd have it's own method of sorting out the change to non-associations. but if we change ours and they change theirs, how on earth can we align the sync?

xoen: also, something to bear in mind is that we may potentially delete an open non-association but leave a closed one, which may be dangerous.

src/main/resources/db/migration/V1_9__add_unique_constraint.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE non_association
+ADD CONSTRAINT prisoner_association_unq UNIQUE (first_prisoner_number, second_prisoner_number)
```
xoen, line 3: Do we need this unique constraint already? If we allow for (potential) duplicated we don't need to over complicate the merge logic right? (as we can simply replace `oldPrisonerNumber` with `newPrisonerNumber` as you were doing before)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
Mjwillis: The merge will already have happened in NOMIS so not we don't sync
Like in incentive we just need to align to what NOMIS now has.  

I will test out an set of NA in NOMIS and perform a merge to observe what happens.

src/main/resources/db/migration/V1_9__add_unique_constraint.sql
Mjwillis: Oops

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
Mjwillis: These are very JPA additions and are often need when dealing with collections

src/main/resources/db/migration/V1_9__add_unique_constraint.sql
ushkarev: i think we're _forced_ into this situation rather than want it! i would expect that the rest of the team would think that closed non-associations remain in the history forever rather than getting re-opened with new details

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
```diff
@@ -54,26 +49,43 @@ class NonAssociationsMergeService(
...
+  private fun merge(
+    nonAssociation: NonAssociation,
+    duplicateRecord: NonAssociation?,
+    newPrisonerNumber: String,
+    otherPrisonerNumber: String,
+    primary: Boolean,
+  ) =
+    if (duplicateRecord != null) {
+      log.info("Deleting non-association record ${nonAssociation.id} - Duplicate")
+      nonAssociationsRepository.delete(duplicateRecord)
```
xoen, line 78: Did you mean `nonAssociationsRepository.delete(nonAssociation)` here? Sorry if this is correct.

src/main/resources/db/migration/V1_9__add_unique_constraint.sql
Mjwillis: Ok I'll remove ! - Think we need to have a rule about only one open -discuss tomorrow

ushkarev (approved):

PR #35, opened 2023-07-17
NON-88: Get non-associations for a prisoner

xoen:
`GET /prisoner/:prisonerNumber/non-associations` endpoint now returns records from the database.

- service uses OffenderSearch API to hydrate responses with prisoners information (e.g. first name, last name, etc...)
- endpoint responds `404 Not Found` when the given prisoner can't be found
- however, if a prisoner is known (e.g. valid prisonerNumber) but doesn't have any non-associations the response will be a `200 OK`. This will contain the prisoner information but the `nonAssociations` field will be an empty list (because the person doesn't have any non-associations)
- at the moment endpoint returns only open non-associations in the same prison as the prisoner in the request URL (this is a good default)
- ability to change this default filtering via query parameters will be added separately


docker-compose-local.yml
```diff
@@ -16,50 +16,6 @@ services:
       - "${TMPDIR:-/tmp/localstack}:/var/lib/localstack"
       - "/var/run/docker.sock:/var/run/docker.sock"
 
-  hmpps-auth:
```
Mjwillis, line 19: Why did you remove this?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
```diff
@@ -63,16 +64,22 @@ class NonAssociationsResource(
...
+    )
   }
 
   @PostMapping("/non-associations")
   @PreAuthorize("hasRole('ROLE_NON_ASSOCIATIONS') and hasAuthority('SCOPE_write')")
   @ResponseStatus(HttpStatus.CREATED)
   @Operation(
     summary = "Creates a non-association between two prisoners.",
-    description = "Requires ROLE_NON_ASSOCIATIONS role with write scope.",
+    description = "Requires ROLE_NON_ASSOCIATIONS role with write scope. Also requires ROLE_GLOBAL_SEARCH or ROLE_PRISONER_SEARCH role.",
```
Mjwillis, line 82: The API should have its own creds to give it access to these APIs - the user calling this application would not

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/OffenderSearchService.kt
```diff
@@ -0,0 +1,54 @@
...
+  private fun getClient(useClientCredentials: Boolean = false): WebClient {
+    return if (useClientCredentials) offenderSearchClientCredentials else offenderSearchWebClient
+  }
+
+  /**
+   * Search prisoners by their prisoner number
+   *
+   * Requires ROLE_GLOBAL_SEARCH or ROLE_PRISONER_SEARCH role.
+   */
+  fun searchByPrisonerNumbers(
```
Mjwillis, line 27: So would use client creds to call this - the client will be given the right roles

Mjwillis (commented):
Main comment was on the roles part - it can be tweaked later

docker-compose-local.yml
xoen: These are moved in the `docker-compose.yml`, this is the local one that we use to run the API against `dev` environment but running DB/localstack in docker-compose. The `docker-compose.yml` has everything.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -29,11 +39,46 @@ class NonAssociationsService(
...
+  fun getLegacyDetails(prisonerNumber: String): LegacyNonAssociationDetails {
     return prisonApiService.getNonAssociationDetails(prisonerNumber)
   }
 
   private fun persistNonAssociation(nonAssociation: NonAssociationJPA): NonAssociationJPA {
     return nonAssociationsRepository.save(nonAssociation)
   }
 }
+
+data class NonAssociationOptions(
```
ushkarev, line 81: could the name be something like _filter_ options? cos it's not options configuring a non-association itself

xoen: I didn't name it "filter" 'cause I left it open for configuring more things, e.g. sorting, pagination, etc...but I will rename to plural as it makes more sense.

xoen: (and I wouldn't really want to have multiple parameters passed in for different options related to the non-associations you're getting back)

xoen: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/35/commits/7945067d480f0cde81ddbf697f02def473d9455b

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
xoen: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/35/commits/2d6f8b6a6752a1008daad07a4d47555c5df1f2ff

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/OffenderSearchService.kt
xoen: https://github.com/ministryofjustice/hmpps-non-associations-api/pull/35/commits/2d6f8b6a6752a1008daad07a4d47555c5df1f2ff

xoen:
rebased and tweaked as per PR comments

ushkarev (approved):

PR #37, opened 2023-07-21
NON-88: Ability to sort a prisoner non-associations

xoen:
- non-associations returned in order of creation by default (from newer to older)
- optional `sortBy`/`sortDirection` query params to control sorting criteria and direction
- can be sorted by creation time (default), last name or first name of the other prisoner involved in the non-association
- should be trivial to add more sorting options if/when necessary

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
@@ -5,6 +5,7 @@ import io.swagger.v3.oas.annotations.media.Content
 import io.swagger.v3.oas.annotations.media.Schema
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.tags.Tag
+import org.springframework.data.domain.Sort
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.security.access.prepost.PreAuthorize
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
```diff
@@ -83,12 +85,32 @@ class NonAssociationsResource(
...
+
+    @Schema(
+      description = "Sort non-associations by",
+      required = false,
+      defaultValue = "WHEN_CREATED",
+      example = "LAST_NAME",
+      allowableValues = [
+        "WHEN_CREATED",
+        "LAST_NAME",
+        "FIRST_NAME",
```
ushkarev, line 97: i doubt we'd ever use first name, but i'm totes fine having it as an option. is it worth having prisoner number tho?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
xoen: no idea, but we could add easily if we needed to.

xoen: [Ability to sort by prisoner number](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/37/commits/c73fd951eed644ed80afc9843a8ec915457c3f34)

ushkarev (approved):

PR #38, opened 2023-07-21
NON-141 Initial ground work for sync and migration endpoints

Mjwillis:
Added 
 - `/sync` endpoint
 - `/migrate` endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+      id = null,
+      firstPrisonerNumber = firstPrisonerNumber,
+      firstPrisonerReason = firstPrisonerReason,
+      secondPrisonerNumber = secondPrisonerNumber,
+      secondPrisonerReason = secondPrisonerReason,
+      restrictionType = restrictionType,
+      comment = comment ?: "NO COMMENT",
+      authorisedBy = authorisedBy ?: SYSTEM_USERNAME,
+    )
+  }
```
xoen, line 50: Watch out when creating a new instance of `NonAssociation`, this is missing the close-related fields which means they'll default to an open non-association.

It's fine when creating a new, open, non-association but when syncing existing non-associations it's possible some of them will be closed and we don't want to accidentally re-open them.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+  fun toNewEntity(): NonAssociation {
+    return NonAssociation(
+      id = null,
+      firstPrisonerNumber = firstPrisonerNumber,
+      firstPrisonerReason = firstPrisonerReason,
+      secondPrisonerNumber = secondPrisonerNumber,
+      secondPrisonerReason = secondPrisonerReason,
+      restrictionType = restrictionType,
+      comment = comment ?: "NO COMMENT",
+      authorisedBy = authorisedBy ?: SYSTEM_USERNAME,
```
xoen, line 48: Should this field be whatever it was in NOMIS, e.g. retain the information about the person who created the original non-association?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+) {
+  fun toNewEntity(): NonAssociation {
+    return NonAssociation(
+      id = null,
+      firstPrisonerNumber = firstPrisonerNumber,
+      firstPrisonerReason = firstPrisonerReason,
+      secondPrisonerNumber = secondPrisonerNumber,
+      secondPrisonerReason = secondPrisonerReason,
+      restrictionType = restrictionType,
+      comment = comment ?: "NO COMMENT",
```
xoen, line 47: I'd probably leave as empty string as it's more neutral and potentially leave the UI the ability to detect this and possibly show whatever message content decide (and can be changed).

That would be my preference as it's more neutral, but not a biggie tho :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
xoen: Actually nevermind, this is falling back to the system username so it's fine as long as we're happy for migration to not have to pass this field :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/SyncRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+      id = null,
+      firstPrisonerNumber = firstPrisonerNumber,
+      firstPrisonerReason = firstPrisonerReason,
+      secondPrisonerNumber = secondPrisonerNumber,
+      secondPrisonerReason = secondPrisonerReason,
+      restrictionType = restrictionType,
+      comment = comment ?: "NO COMMENT",
+      authorisedBy = authorisedBy ?: SYSTEM_USERNAME,
+    )
+  }
```
xoen, line 50: [Same comment](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/38/files#r1270666533) to the corresponding migration method: Watch out because this would create an open non-association which may not be what we want if the non-association synced from NOMIS is actually closed.

This is _possibly_ fine for syncing _new_ non-associations, but I would be more explicit nevertheless, something to bear in mind anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+    required = true,
+    example = "PERPETRATOR",
+  )
+  val secondPrisonerReason: NonAssociationReason,
+
+  @Schema(description = "Type of restriction, e.g. don't locate in the same cell", required = true, example = "CELL")
+  val restrictionType: NonAssociationRestrictionType,
+
+  @Schema(
+    description = "Type of restriction, e.g. don't locate in the same cell",
```
xoen, line 30: the description probably needs tweaking to something like `"Explanation of why prisoners are non-associated"`, given it's for the comment field :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/SyncRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+    required = true,
+    example = "PERPETRATOR",
+  )
+  val secondPrisonerReason: NonAssociationReason,
+
+  @Schema(description = "Type of restriction, e.g. don't locate in the same cell", required = true, example = "CELL")
+  val restrictionType: NonAssociationRestrictionType,
+
+  @Schema(
+    description = "Type of restriction, e.g. don't locate in the same cell",
```
xoen, line 30: again, swagger docs description is incorrect for the comment field :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncResource.kt
```diff
@@ -0,0 +1,60 @@
...
+      ApiResponse(
+        responseCode = "403",
+        description = "Missing required role. Requires the ROLE_NON_ASSOCIATIONS_SYNC role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun sync(
+    @Valid @RequestBody
+    syncRequest: SyncRequest,
```
xoen, line 56: I'd probably rename the request DTO class to _something like_ `SyncCreateRequest` as I'd imagine we'll have other endpoints to update (delete too? 🤷 ) non-associations

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncAndMigrateResourceTest.kt
```diff
@@ -0,0 +1,264 @@
...
+          "firstPrisonerReason": "${request.firstPrisonerReason}",
+          "secondPrisonerNumber": "${request.secondPrisonerNumber}",
+          "secondPrisonerReason": "${request.secondPrisonerReason}",
+          "restrictionType": "${request.restrictionType}",
+          "comment": "${request.comment}",
+          "authorisedBy": "$SYSTEM_USERNAME",
+          "isClosed": false,
+          "closedReason": null,
+          "closedBy": null,
+          "closedAt": null
```
xoen, line 122: Is the assumption here that all non-associations migrated via this endpoint will be open? Aren't we going to migrate existing closed non-associations too?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncAndMigrateResourceTest.kt
```diff
@@ -0,0 +1,264 @@
...
+          "firstPrisonerReason": "${request.firstPrisonerReason}",
+          "secondPrisonerNumber": "${request.secondPrisonerNumber}",
+          "secondPrisonerReason": "${request.secondPrisonerReason}",
+          "restrictionType": "${request.restrictionType}",
+          "comment": "NO COMMENT",
+          "authorisedBy": "$SYSTEM_USERNAME",
+          "isClosed": false,
+          "closedReason": null,
+          "closedBy": null,
+          "closedAt": null
```
xoen, line 246: Similarly, is the assumption that new non-associations are created open when synced with our service?

Mjwillis: Yes all valid points - I'll update

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncAndMigrateResourceTest.kt
Mjwillis: yes we are

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
```diff
@@ -27,14 +28,23 @@ data class MigrateRequest(
...
     required = false,
     example = "John and Luke always end up fighting",
   )
   val comment: String? = null,
 
   @Schema(description = "Who authorised the non-association", required = false, example = "John Smith")
   val authorisedBy: String? = null,
+
+  @Schema(description = "Indicates that the NA is active", required = true, example = "false")
+  val active: Boolean,
```
xoen, line 41: Could this be `isClosed` for consistency with the rest? I'm sure Syscon can easily map `active` to `isClosed` :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SyncAndMigrateService.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.CreateSyncRequest
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.MigrateRequest
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociation
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.UpdateSyncRequest
+import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.repository.NonAssociationsRepository
+
+@Service
+@Transactional
+class SyncAndMigrateService(
```
ushkarev, line 17: can i double check, is the purpose of having a bespoke sync/migrate service, rather than reusing the normal service, basically to not publish events that would create an infinite sync loop?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/MigrateRequest.kt
Mjwillis: No, I want to keep the contract matching the data in NOMIS as these endpoint will be delete one live and migrated

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/SyncRequest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+  @Schema(description = "Type of restriction, e.g. don't locate in the same cell", required = true, example = "CELL")
+  val restrictionType: NonAssociationRestrictionType,
+
+  @Schema(
+    description = "Explanation of why prisoners are non-associated",
+    required = false,
+    example = "John and Luke always end up fighting",
+  )
+  val comment: String? = null,
```
ushkarev, line 35: why is the comment optional? does NOMIS make a distinction between empty and null here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SyncAndMigrateService.kt
Mjwillis: Yes and also we can delete this code afterwards

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/SyncRequest.kt
Mjwillis: Its a nullable field in NOMIS

ushkarev (approved):

PR #42, opened 2023-07-27
Serialise dates properly

ushkarev:


src/main/resources/application.yml
```diff
@@ -26,9 +26,12 @@ spring:
             token-uri: ${api.base.url.oauth}/oauth/token
 
   jackson:
-    date-format: "yyyy-MM-dd HH:mm:ss"
```
ushkarev, line 29: this is not how we want dates formatted. in fact, this is not even how they _were_ formatted because this setting only applies to `java.util.Date`.

src/main/resources/application.yml
```diff
@@ -26,9 +26,12 @@ spring:
...
+    date-format: "yyyy-MM-dd'T'HH:mm:ssZ"
     serialization:
-      WRITE_DATES_AS_TIMESTAMPS: false
+      write-dates-as-timestamps: false
+      write-dates-with-context-time-zone: true
+      write-dates-with-zone-id: true
+    time-zone: "Europe/London"
```
ushkarev, line 34: i think we should output time zone whenever possible

PR #43, opened 2023-07-28
NON-149 Provide Close Non-association API endpoint.

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -72,22 +78,30 @@ class NonAssociationsService(
       "Non-association with ID $id not found",
     )
 
-    val updatedNonAssociation = nonAssociationsRepository.save(
-      nonAssociation.updateWith(update),
-    )
+    nonAssociation.updateWith(update)
 
     log.info("Updated Non-association [$id]")
-    telemetryClient.trackEvent(
```
Mjwillis, line 80: This is logged already in audit

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -72,22 +78,30 @@ class NonAssociationsService(
       "Non-association with ID $id not found",
     )
 
-    val updatedNonAssociation = nonAssociationsRepository.save(
```
Mjwillis, line 75: Don't need to save

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/NonAssociationsRepositoryTest.kt
```diff
@@ -72,7 +72,7 @@ class NonAssociationsRepositoryTest : TestBase() {
 
     val closedBy = "Aldo"
     val closedReason = "They're friends now"
-    val closedAt = LocalDateTime.now()
+    val closedAt = LocalDateTime.now(clock)
```
ushkarev, line 75: i started feeding in `clock` into a bunch of places in tests but then realised that `whenCreated` comes from spring wizardry so will actually be the real world "now". it would be nice if we can teach *spring* to use a fixed clock during tests but i didn't dig deep enough

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
```diff
@@ -100,6 +103,19 @@ data class PatchNonAssociationRequest(
...
 
+/**
+ * Request format to close a non-association between two prisoners
+ */
+@Schema(description = "Request to close a non-association")
+data class CloseNonAssociationRequest(
+  @Schema(description = "Reason for closing the non-association", required = true, example = "They are friends now")
+  val closureReason: String,
+  @Schema(description = "Date and time of the closure, if not provided will default to today's time", required = false, example = "2023-06-07", defaultValue = "now")
+  val dateOfClosure: LocalDateTime? = null,
```
ushkarev, line 114: why would we want clients of the api to specify any time other than "now"? for sync/migrate, sure, we need to consider other times, but for this…?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
```diff
@@ -137,24 +138,75 @@ class HmppsNonAssociationsApiExceptionHandler {
...
+/**
+ * Codes that can be used by api clients to uniquely discriminate between error types,
+ * instead of relying on non-constant text descriptions.
+ *
+ * NB: Once defined, the values must not be changed
+ */
+enum class ErrorCode(val errorCode: Int) {
+  NonAssociationAlreadyClosed(100),
+  UserInContextMissing(401),
+}
```
ushkarev, line 185: 👍 LIKE

ushkarev (approved):
generally happy to left some comments

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
Mjwillis: I think its useful for potential data corrections

PR #46, opened 2023-07-30
NON-154 Add Cell ID back in as used in cell move

Mjwillis:
Will need to call `/api/locations/code/{code}` in prison API for now to get ID when feature switch is active to not use NOMIS data

Future task to bring cell move to use description rather than ID as location - offending line is https://github.com/ministryofjustice/cell-moves/blob/57cf273f503403fb51933ccdf68555a18bb4821f/backend/controllers/cellMove/selectCell.ts#L61


Mjwillis:
Actually, don't think it is used after all

PR #48, opened 2023-07-31
NON-148: Expose user who last updated a non-association in the payload

ushkarev:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResourceTest.kt
```diff
@@ -27,13 +26,16 @@ import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.NonAssociation as NonAssociationJPA
 
-@WithMockUser
+// language=text
+const val expectedUsername = "A_TEST_USER"
+
+@WithMockUser(username = expectedUsername)
```
ushkarev, line 32: the auth façade uses this username instead of what's passed in as headers in the request when running tests. the default was `"user"`

Mjwillis (approved):
LGTM
Think the authorised by field is not an audit field in the same way as who update the last record

PR #51, opened 2023-08-01
NON-151: Return number of open/closed non-associations

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/NonAssociationsRepository.kt
```diff
@@ -6,7 +6,12 @@ import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.NonAssociation
...
-  fun findAllBySecondPrisonerNumber(prisonerNumber: String): List<NonAssociation>
+  fun findAllByFirstPrisonerNumberOrSecondPrisonerNumber(firstPrisonerNumber: String, secondPrisonerNumber: String): List<NonAssociation>
   fun findByFirstPrisonerNumberAndSecondPrisonerNumber(firstPrisonerNumber: String, secondPrisonerNumber: String): NonAssociation?
 }
+
+/**
+ * Where the given prisoner number is either first OR second in the non-association
+ */
+fun NonAssociationsRepository.findByPrisonerNumber(prisonerNumber: String): List<NonAssociation> =
+  findAllByFirstPrisonerNumberOrSecondPrisonerNumber(prisonerNumber, prisonerNumber)
```
ushkarev, line 17: fyi: needs to be an extension function rather than a concrete method inside the interface because spring still tries to turn it into a materialised method otherwise

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/NonAssociationsRepository.kt
```diff
@@ -6,7 +6,27 @@ import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.NonAssociation
...
+/**
+ * Returns all non-associations given two prisoner numbers
+ */
+fun NonAssociationsRepository.findAllByPairOfPrisonerNumber(prisonerNumbers: Pair<String, String>): List<NonAssociation> =
+  findAllByFirstPrisonerNumberAndSecondPrisonerNumberOrFirstPrisonerNumberAndSecondPrisonerNumber(
+    prisonerNumbers.first,
+    prisonerNumbers.second,
+    prisonerNumbers.second,
+    prisonerNumbers.first,
+  )
```
ushkarev, line 32: fyi: these need to be extension functions rather than concrete methods inside the interface. this is because spring still tries to turn them into materialised methods even if they're not abstract.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
```diff
@@ -216,6 +227,83 @@ class NonAssociationsResource(
...
+    @RequestParam(required = false)
+    secondPrisonerNumber: String?,
+
+    @Schema(
+      description = "Whether to include open non-associations or not",
+      required = false,
+      defaultValue = "true",
+      example = "false",
+    )
+    @RequestParam(required = false, defaultValue = "true")
```
Mjwillis, line 273: Shouldn't these be required = true

ushkarev: `firstPrisonerNumber` and `secondPrisonerNumber` are technically required (this is reflected in swagger docs) but if you make them required in the `RequestParam` annotation, then you get an unhelpful 500 internal server error if they're absent. hence the manual check in the method body that returns a 400 instead since the _user_ request wasn't valid.

`includeClosed` and `includeOpen` are genuinely not required since there's a sensible default we can use

PR #59, opened 2023-08-14
NON-159 Provide initial delete API endpoint for backend use only

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
```diff
@@ -121,6 +121,17 @@ data class CloseNonAssociationRequest(
...
 
+/**
+ * Request format to delete a non-association between two prisoners
+ */
+@Schema(description = "Request to delete a non-association")
+data class DeleteNonAssociationRequest(
+  @Schema(description = "Reason for deleting the non-association", required = true, example = "Created in error and removed on requested from OMU team")
+  val deletionReason: String,
+  @Schema(description = "The name of the member of staff requesting the deletion", required = true, example = "Andrew Jones")
+  val staffMemberRequestingDeletion: String,
```
ushkarev, line 132: should we make this a username so that it matches the way existing fields are persisted?

ushkarev (approved):

PR #63, opened 2023-08-22
NON-194: Include codes and descriptions in all ’modern’ API responses

ushkarev:
There was an inconsistency in the data model of non-association responses:
• endpoints returning single non-associations only returned internal codes for roles, reasons and restriction types
• endpoint returning a _list_ of many non-associations also returned the human-readable descriptions for each of these

Now:
• all endpoints return codes _and_ descriptions for roles, reasons and restriction types
• BUT NOTE: the endpoint returning a list of many non-associations has had fields renamed to match the many other endpoints, request payloads and entities! The `Code` suffix has been dropped.


doc/architecture/decisions/0002-NOMIS-na-schema.md
```diff
@@ -14,7 +14,7 @@ This illustrates the key fields that NOMIS holds on non-associations
 
 ```mermaid
 ---
-title: Non Associations NOMIS ER Diagram
+title: Non-associations NOMIS ER Diagram
```
xoen, line 17: 🏆

xoen (approved):

PR #65, opened 2023-08-23
INC-201 Change NOMIS to DPS sync from insert/update to upsert

Mjwillis:
Rules:
- If no NA exists for this prisoner combination a new record is created
- If an active NA exists for this prisoner combination then this record is updated
- If no active and one or more non-active NA exists the the latest (by updated Date) is updated
- on a delete request ALL records for this non-association for these 2 prisoners are removed (active and non-active)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncResource.kt
```diff
@@ -116,10 +84,9 @@ class SyncResource(
       ),
     ],
   )
-  fun syncDelete(
-    @Schema(description = "Non-association ID", example = "1234123423", required = true)
-    @PathVariable
-    id: String,
+  fun sync(
```
xoen, line 87: Isn't this complaining that the 2 functions are the same? (I'm looking this in GH so maybe I'm missing something)

Hang on, this is using method overloading or something isn't it? I see 👌

Mjwillis: yes overloading

ushkarev (approved):

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SyncAndMigrateService.kt
```diff
@@ -25,74 +28,92 @@ class SyncAndMigrateService(
...
-  }
+  fun sync(syncRequest: UpsertSyncRequest): NonAssociation {
+    val prisonersToKeepApart = listOf(
+      syncRequest.firstPrisonerNumber,
+      syncRequest.secondPrisonerNumber,
+    )
+    val existingRecords =
+      nonAssociationsRepository.findAnyBetweenPrisonerNumbers(prisonersToKeepApart, NonAssociationListInclusion.ALL)
+    val latestClosedRecord = existingRecords.filter { na -> na.isClosed }.maxByOrNull { na -> na.whenUpdated }
+    val recordToUpdate = existingRecords.firstOrNull { na -> na.isOpen } ?: latestClosedRecord
```
xoen, line 39: I may miss something but there seems to be an assumption that `existingRecords` are sorted from new to old (by whenUpdated or something) - but I don't think this is the case?

It would mean that if order is arbitrary an older closed non-association will be selected for update instead of potentially a newer, open, non-association.

I suppose it's a bit of an edge case but maybe it would be safer to sort all the non-association by `whenUpdated` so that newest open is updated (or newest closed if they're all closed)

It really depends on what's the desired/indented behaviour:
1. Do we want to update the most recently updated non-association regardless from whether it's open/closed (in that case no point in checking if it's closed)
2. Or do we want to update the most recently open non-association if any, otherwise fallback to the most recently modified closed one?

2 seems more sensible and I think it's in line with what we want? We'll probably have to tweak the logic separately, but I'll finish have a look at this PR in case there is something else that explain this.

xoen: I've played with this locally and AFAICT the logic is actually fine as the `existingRecords.firstOrNull { na -> na.isOpen }` would only potentially return the first _**open**_ non-association it finds.

Although the `existingRecords` are not sorted by `whenCreated DESC` the fact there should always be at max one single open non-association means it should be causing a problem in this case.

Apologies for the noise, resolving this 👍

PR #75, opened 2023-08-31
NON-215: Ability to filter non-associations by prison for "between"/"involving" endpoints

xoen:
- optional `prisonId` query param
- when provided information about the prisoners are retrieved and only non-associations between prisoners which are both at the given prison are returned
- response format is unchanged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
@@ -90,24 +90,44 @@ class NonAssociationsService(
 
   /**
    * Returns all non-associations that exist amongst provided group of prisoners
+   *
+   * Optionally a prisonId can be provided to only return non-associations
+   * where both prisoners are at this given prison
    */
   fun getAnyBetween(
     prisonerNumbers: Collection<String>,
     inclusion: NonAssociationListInclusion = NonAssociationListInclusion.OPEN_ONLY,
+    prisonId: String? = null,
   ): List<NonAssociationDTO> {
-    return nonAssociationsRepository.findAnyBetweenPrisonerNumbers(prisonerNumbers, inclusion)
-      .map { it.toDto() }
+    var nonAssociations = nonAssociationsRepository.findAnyBetweenPrisonerNumbers(prisonerNumbers, inclusion)
+
+    // return only non-associations where both prisoners are in given prison
+    if (prisonId != null) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -216,6 +236,18 @@ class NonAssociationsService(
     }
   }
 
+  private fun filterByPrisonId(nonAssociations: List<NonAssociationJPA>, prisonId: String): List<NonAssociationJPA> {
+    val prisonerNumbers = nonAssociations.map { nonna ->
+      listOf(nonna.firstPrisonerNumber, nonna.secondPrisonerNumber)
+    }.flatten()
```
ushkarev, line 242: i'd use a set

Mjwillis:
@xoen I think this endpoint is not what the appointments team asked for.  

They just want a list of all non-associations for each prisoner included in a given list.  
Yes they can filter by active / inactive but I think the prison filtering is less important.  

The upshot is that a non-association should still be returned even if not in the list - if they have one with that prisoner.

Mjwillis (commented):
Nothing wrong with this change per se but I don't think it's what they want unfortuately.

ushkarev (approved):

Mjwillis (approved):
Happy to approve but see other comment

xoen:
>They just want a list of all non-associations for each prisoner included in a given list.
> [...]
> The upshot is that a non-association should still be returned even if not in the list - if they have one with that prisoner.

@Mjwillis that's what the `POST /non-associations/involving` endpoint does (logic was added [in this other PR](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/71/files))
For example, if you say give me non-associations for prisoners A, B and C all their non-associations will be returned (even if the other prison is not in the provided list), e.g.
- (A, B) and (A, Z) will be returned (because at least one of the prisoners provided is involved)
- (X, Y) or (Z, D) will **not** (because none of the prisoners in these non-associations were provided)

Hope that makes sense ☺️

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
xoen: As discussed offline, I'm tweaking this to:
- use `flatMap()` to do collection/flattening in one go (e.g. removing the final `flatten()` call)
- make prisonerNumbers a `Set` just before sending the request to Offender Search API

PR with these tweaks coming soon

PR #78, opened 2023-09-01
NON-215: Don't send duplicated prisoner numbers to Offender Search API

xoen:
Mainly that plus a couple of other tweaks:
- use `flatMap()` instead of `map()` + `flatten()`
- added a note in `getPrisonerNonAssociations()` on why in that method we use `flatMapTo()` instead of `flatMap()` to be sure the "key" prisoner's prisoner number is always in the request to Offender Search API (a test covers that)


ushkarev (approved):

ushkarev:
✨

PR #79, opened 2023-09-02
NON-193: Allow sorting non-associations by more options

ushkarev:
Even though the staff-facing site will have to sort non-association lists internally, we might as well have the same options in the api to prevent confusion.


Mjwillis (approved):
LGTM

PR #89, opened 2023-09-08
NON-225 Validate and up char to 60

Mjwillis:


src/main/resources/db/migration/V1_2__authorised_by_size_change.sql
```diff
@@ -0,0 +1,2 @@
+alter table non_association
+    alter column authorised_by type varchar(60);
```
ushkarev, line 2: We should do this on all username/staff name fields if we do it here, no?

xoen, line 2: If we increase this column here because usernames may be longer we should also increase the capacity of `updated_by` and `closed_by` as well as they may have the same problem as they are field used to store usernames as well.

xoen (commented):
Looks good to me but I think the other columns used to store usernames should be increased in capacity as well as they may have similar problems later on.

src/main/resources/db/migration/V1_2__authorised_by_size_change.sql
Mjwillis: No, because they are user names from NOMIS which are 30 chars

Authorised by is a free text field in NOMIS of 60 chars

Mjwillis: Nope! @xoen see below

Mjwillis:
@xoen ok I don't think they can be happy to increase in future PR - like to get merge to fix sync issue first

Mjwillis:
Hold off for a minute

PR #90, opened 2023-09-08
NON-222 Record the source of where event comes from so it can be ignored by DPS->NOMIS sync

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/SyncResource.kt
```diff
@@ -86,11 +92,19 @@ class SyncResource(
...
+    deleteRequest: DeleteSyncRequest,
   ) =
-    syncAndMigrateService.delete(syncRequest)
+    syncAndMigrateService.delete(deleteRequest)
+      .onEach { deletedNonAssociation ->
+        eventPublishService.publishEvent(
+          NonAssociationDomainEventType.NON_ASSOCIATION_DELETED,
+          deletedNonAssociation,
+          deleteRequest,
+          "nomis",
```
steverendell, line 105: I could go with this , but in other syncs / migrations we have used values 'NOMIS' and 'DPS'

xoen (approved):
Looks good but I agree with @steverendell that it would be best if the source values are consistent with other sync/migration services and possibly use an enum for the field.

PR #108, opened 2023-09-20
NON-230: List of non associations filtering

ushkarev:
… so that the new endpoint is more similar to existing ones

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
@@ -543,6 +543,11 @@ class NonAssociationsResource(
         description = "A page of Non-associations are returned",
       ),
       ApiResponse(
+        responseCode = "400",
+        description = "When pagination parameters are not valid",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/resource/NonAssociationsResource.kt
```diff
@@ -196,10 +196,71 @@ class NonAssociationsResource(
...
+    @RequestParam(required = false, defaultValue = "true")
+    includeOpen: Boolean = true,
+
+    @Schema(
+      description = "Whether to include closed non-associations or not",
+      required = false,
+      defaultValue = "true",
+      example = "false",
+    )
+    @RequestParam(required = false, defaultValue = "true")
```
xoen, line 249: I don't know how I feel about `includeClosed` by default just for this endpoint. It's inconsistent with other endpoints which can potentially lead to confusion.

I understand that:
1. this endpoint was originally return everything
2. this will be likely used mainly by Syscon for recoinciliation

But I still think if this same option is usually false by default (e.g. only open non-associations are returned by default) we should do the same in this endpoint.

ushkarev: lol how _strongly_ do you feel? i don't mind flipping this, but there's already a consumer of the endpoint…

xoen (approved):
❤️

PR #111, opened 2023-09-27
Improve documentation and normalise imports in error handling

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
```diff
@@ -7,7 +7,10 @@ import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.CONFLICT
+import org.springframework.http.HttpStatus.FORBIDDEN
 import org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR
+import org.springframework.http.HttpStatus.NOT_FOUND
+import org.springframework.http.HttpStatus.UNAUTHORIZED
```
ushkarev, line 13: half of these were used as `HttpStatus.XXX` the other half imported directly as `XXX` – might as well use the latter method consistently 🤷

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
```diff
@@ -260,6 +263,6 @@ class NonAssociationAlreadyClosedException(id: Long) : Exception("Non-associatio
 
 class UserInContextMissingException : Exception("There is no user in context for this request")
 
-class OpenNonAssociationAlreadyExistsException(prisoners: List<String>) : Exception("Prisoners [$prisoners] already have open non-associations")
+class OpenNonAssociationAlreadyExistsException(prisoners: List<String>) : Exception("Prisoners $prisoners already have open non-associations")
```
ushkarev, line 266: `$prisoners` already formats as a list surrounded by square brackets

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
xoen: Interesting as I would have probably preferred the other way around and have a single import 😝 (e.g. use `HttpStatus.NOT_FOUND`, `HttpStatus.UNAUTHORIZED`, etc...) 

But then again, doesn't really matter.

xoen: (but maybe from the context is very clear these are HTTP Statuses so it's probably cleaner the way you did it 👍)

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
ushkarev: me too actually. but there were more instances in the file of the second form so i went with the more popular form 🤷

PR #120, opened 2023-10-03
NON-262: Legacy endpoint using DB in production

xoen:
Switched feature flag in production so that `/legacy` endpoint is now using DB instead of proxying to Prison API (which uses NOMIS data)

This is used in [this `when` statement](https://github.com/ministryofjustice/hmpps-non-associations-api/blob/671d2a9df9a2b97234161474752dec2dbd83da42/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt#L244-L253):

```Kotlin
return when (featureFlagsConfig.legacyEndpointNomisSourceOfTruth) {
  true -> prisonApiService.getNonAssociationDetails(prisonerNumber, currentPrisonOnly, excludeInactive)
  false -> {
    val inclusion = if (excludeInactive) NonAssociationListInclusion.OPEN_ONLY else NonAssociationListInclusion.ALL
    return getPrisonerNonAssociations(
      prisonerNumber,
      NonAssociationListOptions(inclusion = inclusion, includeOtherPrisons = !currentPrisonOnly),
    ).toLegacy()
  }
}
```

**NOTE**: Wait for migration/syncing, etc before actually merging/deploying - anything else to consider before merging this?

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -9,7 +9,7 @@ generic-service:
     API_BASE_URL_OAUTH: https://sign-in.hmpps.service.justice.gov.uk/auth
     API_BASE_URL_PRISON: https://api.prison.service.justice.gov.uk
     API_BASE_URL_OFFENDER_SEARCH: https://prisoner-offender-search.prison.service.justice.gov.uk
-    FEATURE_LEGACY_ENDPOINT_NOMIS_SOURCE_OF_TRUTH: true
+    FEATURE_LEGACY_ENDPOINT_NOMIS_SOURCE_OF_TRUTH: false
 
   postgresDatabaseRestore:
     enabled: true


Mjwillis (approved):
LGTM

ushkarev:
NB: the `getDetailsFromPrisonApiByPrisonerNumber` method needs this annotation `@PreAuthorize("hasRole('ROLE_READ_NON_ASSOCIATIONS')")`

xoen:
> NB: the getDetailsFromPrisonApiByPrisonerNumber method needs this annotation @PreAuthorize("hasRole('ROLE_READ_NON_ASSOCIATIONS')")
> [...]

@ushkarev I've briefly discussed this with @Mjwillis and it's a bad idea to change the permissions of this endpoint because it's used by a bunch of clients now.

Some of them are not using their system client token which means their users would have to be updated to have this role (e.g. their app will break for some users).
Even if some clients are using their system token they may not have that role which means this is not a backward-compatible change as it sounds.

ushkarev (approved):

ushkarev (approved):

ushkarev (approved):

PR #129, opened 2023-10-09
NON-272: Node REST client

ushkarev:
Publish a node REST client to use this api from DPS/HMPPS front-end express applications.
See client’s README file for release and usage information.

The REST client was extracted from [hmpps-non-associations](https://github.com/ministryofjustice/hmpps-non-associations/blob/f2b56f9e527b541b4cb02629fe47c8dae1c729b2/server/data/nonAssociationsApi.ts).


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+updates:
+  - package-ecosystem: npm
+    directory: /clients/node
+    allow:
+      - dependency-type: development
+    ignore:
+      - dependency-type: production
+      - dependency-name: @types/node
+        versions:
+          - ">=19"
```
ushkarev, line 11: we want to update packages used in building and testing, but "production" dependencies should really be set by _clients_ of this library.

clients/node/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@ministryofjustice/hmpps-non-associations-api",
+  "version": "0.0.0",
```
ushkarev, line 3: the versions is set from release tag when publishing

clients/node/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "format": "prettier --write src test",
+    "lint": "eslint . --cache --max-warnings 0",
+    "test": "jest"
+  },
+  "engines": {
+    "node": "18 || 20"
+  },
+  "dependencies": {
+    "agentkeepalive": ">=4",
+    "superagent": ">=8"
```
ushkarev, line 34: direct dependencies are just minimal versions and should be set precisely in clients of this library

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt
```diff
@@ -132,13 +132,14 @@ class HmppsNonAssociationsApiExceptionHandler {
   @ExceptionHandler(ResponseStatusException::class)
   fun handleResponseStatusException(e: ResponseStatusException): ResponseEntity<ErrorResponse?>? {
     log.debug("Response status exception caught: {}", e.message)
+    val reason = e.reason ?: "Unknown error"
```
ushkarev, line 135: might as well send "Unknown error" rather than null so that `ErrorResponse`’s fields do not need to be nullable

clients/node/package.json
```diff
@@ -0,0 +1,59 @@
...
+  ],
+  "scripts": {
+    "build": "tsc",
+    "clean": "rm -rf ./dist/",
+    "format": "prettier --write src test",
+    "lint": "eslint . --cache --max-warnings 0",
+    "test": "jest"
+  },
+  "engines": {
+    "node": "18 || 20"
```
xoen, line 30: Could also be looser? I'd imagine we don't care as it's more of a client project choice, e.g. I'd get anything > 18 would do?

Unless we think newer version of Node may break this package somehow?

ushkarev: i just happen to have added testing on v18 (what we all currently use, except dps) and v20 which is about to go LTS (and i figure we will all move to it shortly)

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 #      - store_artifacts:
 #          path: build/reports/jacoco/test/html
 
-  build-node-client:
+  test-node-client:
```
xoen, line 57: ❤️

xoen (approved):
nice one 👍

clients/node/README.md
```diff
@@ -0,0 +1,121 @@
...
+### GitHub Packages – npm registry
+
+Unfortunately, GitHub Packages requires authentication to pull from public npm registries,
+so the setup here is more complex.
+
+In the application repository, create `.npmrc` with:
+
+```text
+registry=https://registry.npmjs.org/
+@ministryofjustice:registry=https://npm.pkg.github.com
```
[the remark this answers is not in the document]
xoen: I was about to say, I wonder if there is a way to make this more specific as I'd imagine there may be other MoJ (`@ministryofjustice/`) node packages published to the "normal" NPM package.

But weird this doesn't break those.

PR #148, opened 2023-10-23
Note the relaxed caseload restrictions for modifying non-associations in the node REST client read-me docs

ushkarev:
Depends on [UI#182](https://github.com/ministryofjustice/hmpps-non-associations/pull/182)


Mjwillis (approved):
LGTM

PR #152, opened 2023-10-25
NON-287: Disallow adding a non-associations to prisoners with null locations

ushkarev:
The api now responds with 409 CONFLICT when trying to create a non-association with a prisoner with a null location.

Also, allow for offender search results that have no prison name or ID. Because those fields _can_ be null, we are forced to propagate the nullability into _this_ api.

**It does mean that our api contract changes!**

Related: [ui#184](https://github.com/ministryofjustice/hmpps-non-associations/pull/184) & [ui#185](https://github.com/ministryofjustice/hmpps-non-associations/pull/185)

### DIFF
--- a/clients/node/README.md
+++ b/clients/node/README.md
@@ -110,6 +110,7 @@ General notes regarding permissions and roles:
 - Users also having the `GLOBAL_SEARCH` role can also _add_, _update_ and _close_ non-associations for prisoners in transfer and where one prisoner is not in a prison that’s not in their caseloads
 - Users also having the `INACTIVE_BOOKINGS` role can also _add_, _update_ and _close_ non-associations for prisoners outside any establishment / released
 - Users must _close_ rather than _delete_ non-associations
+- No users should be able to _add_, _update_ or _close_ non-associations for prisoners without a booking / with a null location
 
 Release a new version
 ---------------------
--- a/clients/node/src/errorTypes.ts
+++ b/clients/node/src/errorTypes.ts
@@ -2,7 +2,7 @@
  * Structure representing an error response from the api, wrapped in SanitisedError.
  *
  * Defined in uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.ErrorResponse class
- * https://github.com/ministryofjustice/hmpps-non-associations-api/blob/f6002aa1da50b8c4ccd3613e970327d5c67c44ae/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt#L240-L261
+ * see https://github.com/ministryofjustice/hmpps-non-associations-api
  */
 export interface ErrorResponse {
   status: number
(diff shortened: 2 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/offendersearch/OffenderSearchPrisoner.kt
```diff
@@ -5,7 +5,7 @@ data class OffenderSearchPrisoner(
   val firstName: String,
   val lastName: String,
 
-  val prisonId: String,
-  val prisonName: String,
+  val prisonId: String?,
+  val prisonName: String?,
   val cellLocation: String?,
```
ushkarev, line 10: this is the main change: all location-type fields are nullable. the rest of the changes in this commit propagate these nulls

Mjwillis (approved):
LGTM

clients/node/README.md
```diff
@@ -110,6 +110,7 @@ General notes regarding permissions and roles:
 - Users also having the `GLOBAL_SEARCH` role can also _add_, _update_ and _close_ non-associations for prisoners in transfer and where one prisoner is not in a prison that’s not in their caseloads
 - Users also having the `INACTIVE_BOOKINGS` role can also _add_, _update_ and _close_ non-associations for prisoners outside any establishment / released
 - Users must _close_ rather than _delete_ non-associations
+- No users should be able to _add_, _update_ or _close_ non-associations for prisoners without a booking / with a null location
```
xoen, line 113: Why tho? A user knows what's doing no? They searched for this individual and they decided to add this person to this non-association.

It seems to me we're trying to be too clever with the logic here. The user probably have more context than the API.

The person may also get a booking soon and the user reasons to add a non-association to this person may be sound?

clients/node/src/errorTypes.ts
```diff
@@ -2,7 +2,7 @@
  * Structure representing an error response from the api, wrapped in SanitisedError.
  *
  * Defined in uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.ErrorResponse class
- * https://github.com/ministryofjustice/hmpps-non-associations-api/blob/f6002aa1da50b8c4ccd3613e970327d5c67c44ae/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/config/HmppsNonAssociationsApiExceptionHandler.kt#L240-L261
+ * see https://github.com/ministryofjustice/hmpps-non-associations-api
```
xoen, line 5: I kinda liked the direct link to the code.
Although sure code change over time, pin the commit hash is an idea but then it may be stale code...so maybe that's why you made the link more generic?

ushkarev: problem is i'm changing `ErrorCode` here so cannot link to the un-merged code… so i thought i might give up on these direct links. what do you think is better 1) drop all direct commit links 2) link to `main` and cross fingers that if something moves, we remember to repoint these links? 3) wait until merge to put in direct commit links

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/offendersearch/OffenderSearchPrisoner.kt
xoen: Wouldn't this have an impact on (other) clients tho? It sounds like a breaking change (although a rare one as I'd imagine most of the data _will_ have the prisoner number)

clients/node/src/errorTypes.ts
xoen: I think this is probably the best compromise: Comment has class path/packages and link is to repo.

xoen (approved):
I'm not convinced we should try to be clever on the API side and prevent creating a non-association just because one of the people involved doesn't have a prisonId.

It should be a rare occasions anyway. But changes look fine.

PS: I think the nullable may be potentially break some of the clients (sometimes). But maybe it's so rare we're ok with this tradeoff?

PR #153, opened 2023-10-27
Use non-standard port to connect to RDS/Redis

xoen:
Default to port `12345` instead of the corresponding standard ports. Especially in the case of RDS, accidentally leave the port forward to a `preprod`/`prod` DB while running tests could be dangerous.


xoen:
@ushkarev does this look OK to you? Can we merge?

ushkarev (approved):

PR #161, opened 2023-11-01
NON-299: Record user creating a non-association using this api

ushkarev:
This is not yet exposed when retrieving individual non-association records, but was already available when listing non-associations for a given prisoner.

The api contract doesn't change, but our record-keeping is improved.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
@@ -85,7 +85,7 @@ data class CreateNonAssociationRequest(
   @field:Size(min = 1, message = "A comment is required")
   val comment: String,
 ) {
-  fun toNewEntity(updatedBy: String, clock: Clock): NonAssociationJPA {
+  fun toNewEntity(createdBy: String, clock: Clock): NonAssociationJPA {
     return NonAssociationJPA(
       id = null,
       firstPrisonerNumber = firstPrisonerNumber,
@@ -95,7 +95,8 @@ data class CreateNonAssociationRequest(
       reason = reason,
       restrictionType = restrictionType,
       comment = comment,
-      updatedBy = updatedBy,
+      authorisedBy = createdBy,
+      updatedBy = createdBy,
       whenCreated = LocalDateTime.now(clock),
       whenUpdated = LocalDateTime.now(clock),
     )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/PrisonerNonAssociations.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/PrisonerNonAssociations.kt
@@ -53,7 +53,7 @@ data class PrisonerNonAssociation(
 
   @Schema(description = "Explanation of why prisoners are non-associated", required = true, example = "John and Luke always end up fighting")
   val comment: String,
-  @Schema(description = "User ID of the person who created the non-association. NOTE: For records migrated from NOMIS/Prison API this is free text and may not be a valid User ID. Additionally, migrated records might use an internal system username", required = true, example = "OFF3_GEN")
+  @Schema(description = "User ID of the person who created the non-association. NOTE: For records migrated from NOMIS/Prison API this is free text and may not be a valid User ID. Additionally, migrated records might use an internal system username. It can be an empty string", required = true, example = "OFF3_GEN")
   val authorisedBy: String,
   @Schema(description = "When the non-association was created", required = true, example = "2021-12-31T12:34:56.789012")
   val whenCreated: LocalDateTime,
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
```diff
@@ -95,7 +95,8 @@ data class CreateNonAssociationRequest(
       reason = reason,
       restrictionType = restrictionType,
       comment = comment,
-      updatedBy = updatedBy,
+      authorisedBy = createdBy,
```
ushkarev, line 98: `authorisedBy` were null when created through this endpoint, but were typically non-null free text when migrated from NOMIS

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/NonAssociation.kt
```diff
@@ -37,7 +37,9 @@ class NonAssociation(
   var restrictionType: RestrictionType,
   var comment: String,
 
-  var authorisedBy: String? = null, // This is stored for legacy reasons - the data is not returned in the API except on the legacy endpoints
```
ushkarev, line 40: > data is not returned in the API except on the legacy endpoints

this is not true. this authorising/creating user is returned when listing all non-associations for a given prisoner in the "modern" api

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
xoen: Would `updatedBy` would be null in this case then? Is that by design?

ushkarev: `updatedBy` is never null

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/PrisonerNonAssociations.kt
```diff
@@ -53,7 +53,7 @@ data class PrisonerNonAssociation(
 
   @Schema(description = "Explanation of why prisoners are non-associated", required = true, example = "John and Luke always end up fighting")
   val comment: String,
-  @Schema(description = "User ID of the person who created the non-association. NOTE: For records migrated from NOMIS/Prison API this is free text and may not be a valid User ID. Additionally, migrated records might use an internal system username", required = true, example = "OFF3_GEN")
+  @Schema(description = "User ID of the person who created the non-association. NOTE: For records migrated from NOMIS/Prison API this is free text and may not be a valid User ID. Additionally, migrated records might use an internal system username. It can be an empty string", required = true, example = "OFF3_GEN")
```
xoen, line 56: `updatedBy` can also be empty string now no? See [other comment](https://github.com/ministryofjustice/hmpps-non-associations-api/pull/161/commits/b48118dd48ec850a8e0cb18238bfa04c0389e361#r1378948269).

I think I would prefer if `updatedBy` was always populated even on creation (would match `authorisedBy`) even tho they didn't _really_ update the record. 

But I don't know what you think and I have no strong opinion on this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
xoen: Tests seem to contradict this, so maybe I'm missing something as it seems like `updatedBy` would match `authorisedBy` on creation as I would expect 🤔

xoen: (to clarify "this" refers to my comment, not yours)

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/PrisonerNonAssociations.kt
xoen: How does `updatedBy` gets populated now when a record is created?

xoen: Alright, ignore me, missed this line for some reason 🤦 

https://github.com/ministryofjustice/hmpps-non-associations-api/pull/161/files#diff-718a1a6e743739af98497c6018f0eb75efaa098dbaa28385e7be95cd23563f16R99

ushkarev: no, `updatedBy` cannot be null. the one place this DTO is created uses `[some nullable field] ?: ""` so it can indeed be a blank string.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/dto/NonAssociation.kt
ushkarev: > it seems like updatedBy would match authorisedBy on creation as I would expect
yes, that's the intention of this PR. previously, `authorisedBy` would always be null when created through the api

PR #184, opened 2023-11-27
NON-318: Use localstack v3 for integration testing

ushkarev:


docker-compose-local.yml
```diff
@@ -2,7 +2,7 @@ version: "3"
 services:
 
   localstack:
-    image: localstack/localstack:2.3
+    image: localstack/localstack:3
```
xoen, line 5: No breaking changes?

ushkarev: read-me said nothing that sounded significant 🤷

PR #214, opened 2024-02-06
NON-351 Remove legacy sync functionality

Mjwillis:
This commit removes legacy sync files that are no longer needed. The deleted files contain configurations and services for syncing non-associations from NOMIS, as well as DTOs for upsert and delete sync requests. This cleanup helps maintain the codebase and reduce complexity.


steverendell (approved):
lgtm

PR #240, opened 2024-03-15
NON-347: Spike to integrate DPR (Digital Prison Reporting) library

Mjwillis:


src/main/resources/definitions.json
```diff
@@ -0,0 +1,268 @@
+{
+  "id": "non-associations",
+  "name": "Non-associations",
+  "description": "Non-associations prison reports",
```
ushkarev, line 4: _prison_ reports? we don’t know anything about which prisons non-associations relate to 😞

ashleygyngell (dismissed):

ushkarev (approved):

src/main/resources/definitions.json
xoen: ☑️ https://github.com/ministryofjustice/hmpps-non-associations-api/pull/240/commits/479a96584bec62f92b6d83202d8f23cd686cae3b

ushkarev (approved):

ushkarev (approved):

PR #255, opened 2024-04-08
NON-358: SAR endpoint responds `204 No Content` when no information matching

xoen:
The endpoint now responds `204 No Content` when no non-associations for the given prisoner were found or when none of them were within the given date range.

**NOTE**: Offender Search API could _potentially_ not have information on some of the prisoners involved in the non-associations found. This is an edge case but without these information (names, etc...) we cannot construct the response correctly so the endpoint will continue to respond `404 Not Found` in these cases.

In my opinion this is better than responding `204 No Content` and misleading the client into believing there were no information on the given subject. It shouldn't happen in theory anyway.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SubjectAccessRequestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SubjectAccessRequestService.kt
@@ -1,7 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.service
 
 import org.springframework.stereotype.Service
-import org.springframework.web.reactive.function.client.WebClientResponseException.NotFound
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.config.SubjectAccessRequestNoContentException
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociationListInclusion
 import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.dto.NonAssociationListOptions
@@ -21,11 +20,7 @@ class SubjectAccessRequestService(
   ): HmppsSubjectAccessRequestContent? {
     val options = NonAssociationListOptions(includeOtherPrisons = true, inclusion = NonAssociationListInclusion.ALL)
 
-    val nonAssociations = try {
-      nonAssociationsService.getPrisonerNonAssociations(prn, options)
-    } catch (e: NotFound) {
-      throw SubjectAccessRequestNoContentException(prn)
-    }
+    val nonAssociations = nonAssociationsService.getPrisonerNonAssociations(prn, options)
 
     // Filter non-associations by given date range
     val content = nonAssociations.copy(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SubjectAccessRequestService.kt
```diff
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.service
 
 import org.springframework.stereotype.Service
+import org.springframework.web.reactive.function.client.WebClientResponseException.NotFound
```
ushkarev, line 4: “reactive” looks suspicious here 👀

xoen: As discussed offline it is what we currently handle in the exception handling file and it works.

I also tried to replace with `import org.springframework.web.client.HttpClientErrorException.NotFound` and it did not work, so it seems like the repository is throwing those exceptions from this package.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/SubjectAccessRequestService.kt
```diff
@@ -18,14 +20,25 @@ class SubjectAccessRequestService(
...
 
-    return HmppsSubjectAccessRequestContent(
-      content = nonAssociations.copy(
-        nonAssociations = nonAssociations.nonAssociations.filter {
-          (fromDate == null || it.whenCreated.toLocalDate().isAfter(fromDate)) && (toDate == null || it.whenCreated.toLocalDate().isBefore(toDate))
-        },
-      ),
+    val nonAssociations = try {
+      nonAssociationsService.getPrisonerNonAssociations(prn, options)
+    } catch (e: NotFound) {
```
Mjwillis, line 26: This doesn't look right to me,  the nonAssociationsService.getPrisonerNonAssociations() has this slightly odd bit:- 
```{kotlin}
 // Throw an exception if any of the prisoners searched were not found
    val missingPrisoners = prisonerNumbers.subtract(foundPrisoners.keys)
    if (missingPrisoners.any()) {
      throw ResponseStatusException(
        HttpStatus.NOT_FOUND,
        "Could not find the following prisoners: $missingPrisoners",
      )
    }
```

Actually this should throw an explicit exception for this use case and the code below would catch that: e..g

```{kotlin}
if (missingPrisoners.any()) {
      throw MissingPrisonersInSearchException( // This is a new exception
        "Could not find the following prisoners: $missingPrisoners",
      )
    }
```
then in this class 
```{kotlin}
// Don't need try catch as exception will bubble up
      nonAssociationsService.getPrisonerNonAssociations(prn, options)
```

Then in Exception Handler 
```{kotlin}
 @ExceptionHandler(MissingPrisonersInSearchException::class)
  fun handleMissingPrisonersInSearchException(e: MissingPrisonersInSearchException): ResponseEntity<ErrorResponse?>? {
    log.debug("Missing prisoners in search: {}", e.message)
    return ResponseEntity
      .status(NOT_FOUND)
      .body(
        ErrorResponse(
          status = NOT_FOUND,
          userMessage = "Missing prisoners: ${e.message}",
          developerMessage = e.message,
        ),
      )
  }
```

Mjwillis (changes requested):
The exception handling needs a tweek

Mjwillis (approved):
LGTM

PR #272, opened 2024-04-30
NON-358: SAR: Respond 209 when a prisoner is not found (in OS API)

xoen:
The Subject Access Request (SAR) endpoint was responding `404 Not Found` when a prisoner number was not found in Offender Search API.

However the SAR endpoint contract doesn't expect 404s: https://non-associations-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/Subject%20Access%20Request/getSarContentByReference

This case is still different than the case where we didn't find non-associations for the subject.

The endpoint now responds `209` "Subject Identifier is not recognised by this service" in this case.


ushkarev (approved):
🤢

PR #311, opened 2024-07-08
NON-318: Sync changes from templates

ushkarev:
Ref:
- [hmpps-ip-allowlists#21](https://github.com/ministryofjustice/hmpps-ip-allowlists/pull/21)
- [hmpps-template-kotlin#215](https://github.com/ministryofjustice/hmpps-template-kotlin/pull/215)
- https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/136#discussion_r1672545276


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsServiceTest.kt
```diff
@@ -29,26 +33,49 @@ import uk.gov.justice.digital.hmpps.hmppsnonassociationsapi.jpa.NonAssociation a
...
     TestBase.clock,
   )
 
+  private fun makeAuthToken(
+    username: String,
+    roles: List<String> = listOf("ROLE_READ_NON_ASSOCIATIONS", "ROLE_WRITE_NON_ASSOCIATIONS"),
+    scopes: List<String> = listOf("read", "write"),
+  ): AuthAwareAuthenticationToken {
+    // adapted from uk.gov.justice.hmpps.test.kotlin.auth.WithMockUserSecurityContextFactory
+    return AuthAwareAuthenticationToken(
```
petergphillips, line 55: if you actually constructed a new `HmppsAuthenticationHolder` instead then you could do the same as [this test](https://github.com/ministryofjustice/prison-api/blob/e11d226f29b1a7ad289d132b7522254c80c1591e/src/test/java/uk/gov/justice/hmpps/prison/service/AppointmentsServiceImplTest.kt#L1212-L1214).  Alternatively I guess you could mock out the authentication token too, or [use spring to wire up](https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/1e5d59bbdc565d5155cbbb60ad464fde12f0ac33/hmpps-kotlin-spring-boot-test-autoconfigure/src/test/kotlin/uk/gov/justice/hmpps/test/kotlin/auth/WithMockAuthUserTest.kt#L9) an instance of the holder.  Final option would be to expose the `makeAuthToken` in the kotlin lib...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsService.kt
```diff
@@ -57,6 +57,9 @@ class NonAssociationsService(
   }
 
   fun createNonAssociation(createNonAssociationRequest: CreateNonAssociationRequest): NonAssociationDTO {
+    val createdBy = authenticationHolder.authenticationOrNull?.userName
```
petergphillips, line 60: if you don't get these exceptions actually being thrown in live, then you could replace by `authenticationHolder.username`.  If you are handling `UserInContextMissingException` then might be better for the `HmppsAuthenticationHolder` to throw a single exception instead of the two it does at present (or rather the two will be subclasses of a single exception) and then you could just let the holder throw its own one

petergphillips: your tests would then be a lot simpler as you could just mock the `username` call on the holder too

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsServiceTest.kt
ushkarev: yeah this bit felt really gross. will see if i can make a concrete `HmppsAuthenticationHolder` because i don't follow the alternatives :)

Mjwillis (approved):
LGTM

PR #377, opened 2024-10-29
NON-369 Prevent calling search when no non-associations are found

Mjwillis:


Mjwillis:
I'll add some tests in another PR

ushkarev:
rather than forbidding calling a few functions with empty collections, i think it's much simpler to just return an empty response from the offender search service. how about this instead (including tests)?
#378

PR #378, opened 2024-10-30
NON-369: Prevent calling search when no non-associations are found

ushkarev:
This should fix the problem of propagating an error from offender search into responses from this api that should have returned empty lists of non-associations


Mjwillis (approved):
LGTM

PR #379, opened 2024-10-31
NON-372 Merge events are failing due to Jackson Dependency upgrade with breaking changes

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/PrisonOffenderEventListener.kt
```diff
@@ -22,6 +22,7 @@ class PrisonOffenderEventListener(
   @SqsListener("nonassociations", factory = "hmppsQueueContainerFactoryProxy")
   @WithSpan(value = "hmpps-non-associations-prisoner-event-queue", kind = SpanKind.SERVER)
   fun onPrisonOffenderEvent(requestJson: String) {
+    log.info("Received event: [$requestJson]")
```
ushkarev, line 25: do we want this to remain given that there is talk on slack about our logging being too costly?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationMergeServiceIntTest.kt
```diff
@@ -21,19 +25,76 @@ import uk.gov.justice.hmpps.test.kotlin.auth.WithMockAuthUser
 
 @DisplayName("Non-associations merge service, integration tests")
 @AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
-@Import(HmppsAuthenticationHolder::class, NonAssociationsMergeService::class, ClockConfiguration::class, ApplicationInsightsConfiguration::class)
+@Import(JacksonAutoConfiguration::class, HmppsAuthenticationHolder::class, NonAssociationsMergeService::class, ClockConfiguration::class, ApplicationInsightsConfiguration::class)
```
ushkarev, line 28: how do we know which test classes this needs to be put on? any others??

build.gradle.kts
```diff
@@ -32,8 +32,6 @@ dependencies {
 
   implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.6.0")
 
-  implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.18.0")
```
Mjwillis, line 35: This is included in hmpps spring boot starter, and this version 2.18 has breaking changes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/PrisonOffenderEventListener.kt
Mjwillis, line 25: Spit out the message BEFORE deserialisation

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationMergeServiceIntTest.kt
```diff
@@ -21,19 +25,76 @@ import uk.gov.justice.hmpps.test.kotlin.auth.WithMockAuthUser
...
+  @Autowired
+  lateinit var mapper: ObjectMapper
+
   @Autowired
   lateinit var service: NonAssociationsMergeService
 
   @Autowired
   lateinit var repository: NonAssociationsRepository
 
+  @Test
```
Mjwillis, line 42: A Test!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/PrisonOffenderEventListener.kt
Mjwillis: Merges are not that frequent and without it had to look for exceptions to find issue

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationMergeServiceIntTest.kt
Mjwillis: No don't think so

build.gradle.kts
ushkarev: it does indeed as we [found out on IRS api](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/198/commits/89daeed7c487cdba983bafd78f7e444de3e34e90), but since tests had passed here we naturally assumed that there was a special thing going on in IRS! i mean surely there was json de/serialisation going on throughout testing…

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationMergeServiceIntTest.kt
ushkarev: how would we know to prevent this sorta thing in future? why is jackson not auto-configured automatically 🤣

Mjwillis: It is now - our spring boot hmpps lib is pin to a working version

build.gradle.kts
ushkarev: but the resource tests deal with json responses and there are plenty of assertions on responses 😕

PR #459, opened 2025-05-20
IR-1053 Upgrade Kotlin and OpenTelemetry dependencies to latest versions

Mjwillis:


build.gradle.kts
```diff
@@ -26,10 +26,10 @@ dependencies {
   implementation("uk.gov.justice.service.hmpps:hmpps-digital-prison-reporting-lib:8.1.0")
 
   implementation("io.opentelemetry:opentelemetry-api:1.50.0")
-  implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.15.0")
+  implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.16.0")
 
   implementation("com.vladmihalcea:hibernate-types-60:2.21.1")
-  implementation("org.hibernate.orm:hibernate-community-dialects:6.6.14.Final")
+  implementation("org.hibernate.orm:hibernate-community-dialects:7.0.0.Final")
```
xoen, line 32: This is a major version change, any reason to think it should break anything or are we in the clear with this?

xoen (dismissed):

build.gradle.kts
ushkarev, line 32: in dependabot's pr #456 there is also 

`testImplementation("io.swagger.parser.v3:swagger-parser:2.1.28")` 

and `kotlin("jvm") version "2.1.21"` in the other build.gradle.kts

Mjwillis: The swagger broke tests so skipped that one for now

ushkarev: the only problem **that i know of** with renovate's automatic prs is that 
`testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.11.4")`
must be kept at that version

ushkarev (approved):

PR #462, opened 2025-06-02
NON-388 Refactor transaction management by removing transactions for read-only endpoints.

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
```diff
@@ -26,7 +26,7 @@ class NonAssociationsMergeService(
 
   /**
    * Replaces an old prisoner number for a new one in all non-associations,
-   * closing newly-duplicate records and deleting non-associations with the same prisoner number on both sides.
+   * closing newly duplicate records and deleting non-associations with the same prisoner number on both sides.
```
ushkarev, line 29: i think it was better with the hyphen 😏
the “newly” attaches to “duplicate” not “records”

ushkarev (approved):

xoen (approved):

PR #486, opened 2025-08-07
Use single principal as token cache key

xoen:
Copied/adapted from: https://github.com/ministryofjustice/hmpps-manage-users-api/pull/319/files

The `ClientCachingOAuth2AuthorizedClientService` has since been updated to work with newer versions of Spring Boot.


Mjwillis (approved):
LGTM

PR #583, opened 2026-04-01
Integrating MoJ SLSA GH Action

xoen:
The aim of this GH Action is to protect from Supply Chain Attacks, e.g. one of your dependency being replaced by a malicious actor in a way or the the other.

The action would check for things like:
- a package being too recent (and therefore potentially having undiscovered vulnerabilities)
- checking for known vulnerabilities (e.g. CVEs)
- checking for typosquatting attacks (e.g. someone releasing a package with a very similar name of a popular one)
- etc...

**NOTE**: This is fairly new and there are still unanswered questions on how it would work in practice. For example if there is a known vulnerability the action will fail the CI run but if a security patch is too new that may also fail the build. It's an ongoing conversation but some of these questions were asked here: https://mojdt.slack.com/archives/C3F8L6XQU/p1775046487179089?thread_ts=1775032049.634239&cid=C3F8L6XQU

See:
- Docs: https://github.com/ministryofjustice/devsecops-actions/tree/main/sca/slsa
- Slack announcement: https://mojdt.slack.com/archives/C3F8L6XQU/p1775032049634239

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -58,6 +58,7 @@ jobs:
     uses: ./.github/workflows/test-node-client.yml
     with:
       node-version: ${{ matrix.node-version }}
+      enable_slsa_action: true
   build:
     name: Build docker image from hmpps-github-actions
     if: github.ref == 'refs/heads/main'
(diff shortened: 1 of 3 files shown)


Mjwillis (approved):
LGTM

PR #599, opened 2026-05-05
IR-1611 NonAssociationsMergeService duplicate detection bugs

Mjwillis:
## Background

When two prisoner records in NOMIS are identified as the same person, a merge event is raised on the SNS queue. `NonAssociationsMergeService` handles this by iterating over all non-associations (NAs) involving the old prisoner number and updating them to the new prisoner number. Where the new prisoner already has an existing NA with the same third party, the migrated record should be closed as a duplicate.

Two bugs were identified in this duplicate-detection logic.

---

### Scenario

- Prisoner **A** (old) has an open NA with third party **C**
- Prisoner **B** (new) has an open NA with third party **C**
- Prisoner **B** also has a **closed** historical NA with **C** from a previous period

### What went wrong

The duplicate lookup used `findByFirstPrisonerNumberAndSecondPrisonerNumber`, a Spring Data JPA derived query that maps to a single-result return type (`NonAssociation?`). When the database contains both an open and a closed record for the same prisoner pair, Spring Data JPA throws:

```
IncorrectResultSizeDataAccessException: Incorrect result size: expected 1, actual 2
```

The merge fails entirely and neither record is updated.

### Root cause

The method was documented with a `TODO` acknowledging this limitation:

```kotlin
// TODO: Assumes that there can only be 1 non-association given a pair of prisoner numbers.
//       In future, this will only be true for open non-associations.
fun findByFirstPrisonerNumberAndSecondPrisonerNumber(...): NonAssociation?
```

Closed historical records were not excluded from the lookup.

---

### Scenario

- Prisoner **A** (old) has an open NA stored as `(A, C)` — A is `firstPrisonerNumber`, C is `secondPrisonerNumber`
- Prisoner **B** (new) has an open NA stored as `(C, B)` — C is `firstPrisonerNumber`, B is `secondPrisonerNumber`

### What went wrong

The duplicate lookup was directional. For an `(A, C)` NA being migrated, it searched for a record where `firstPrisonerNumber = B AND secondPrisonerNumber = C`. Because the existing NA for B was stored in the reverse order `(C, B)`, the lookup returned `null` — no duplicate detected.

The service then updated `(A, C)` to `(B, C)` and marked it `MERGED` (open). The original `(C, B)` record was untouched and remained open. After the merge, both `(B, C)` and `(C, B)` existed as open NAs — a duplicate pair left in the system.

### Root cause

The direction a pair is stored depends on which prisoner was nominated as `firstPrisoner` at the time of creation — an arbitrary choice. The duplicate check assumed a fixed direction, which is not guaranteed.

---

### `NonAssociationsRepository.kt` — new JPQL query

A new method `findOpenBetweenPrisoners(p1, p2)` was added that:

1. Only returns **open** records (fixes Bug 1 — closed historical records are ignored)
2. Checks **both storage directions** `(p1, p2)` and `(p2, p1)` (fixes Bug 2 — direction of storage is irrelevant)

```kotlin
@Query("""
    SELECT n FROM NonAssociation n
    WHERE n.isClosed = false AND (
        (n.firstPrisonerNumber = :p1 AND n.secondPrisonerNumber = :p2) OR
        (n.firstPrisonerNumber = :p2 AND n.secondPrisonerNumber = :p1)
    )
""")
fun findOpenBetweenPrisoners(@Param("p1") p1: String, @Param("p2") p2: String): List<NonAssociation>
```

Returning a `List` rather than a nullable single result also eliminates the risk of a future exception if, for any reason, multiple open records exist for the same pair.

### `NonAssociationsMergeService.kt` — use new query

Both calls to `findByFirstPrisonerNumberAndSecondPrisonerNumber` were replaced with `findOpenBetweenPrisoners(...).firstOrNull()`:

```kotlin
// Before (directional, includes closed records, single-result — throws if > 1 row)
nonAssociationsRepository.findByFirstPrisonerNumberAndSecondPrisonerNumber(
    firstPrisonerNumber = newPrisonerNumber,
    secondPrisonerNumber = nonAssociation.secondPrisonerNumber,
)

// After (bidirectional, open only, safe with multiple historical records)
nonAssociationsRepository.findOpenBetweenPrisoners(
    p1 = newPrisonerNumber,
    p2 = nonAssociation.secondPrisonerNumber,
).firstOrNull()
```

---

### `NonAssociationsMergeServiceTest.kt`

Two regression tests were added to `NonAssociationsMergeServiceTest`:

Verifies that when `findOpenBetweenPrisoners` returns only the open record (as it should, having excluded closed ones), the merge service correctly closes the migrated NA without throwing an exception.

Verifies that when the existing NA for the new prisoner is stored as `(thirdParty, newPrisoner)` rather than `(newPrisoner, thirdParty)`, `findOpenBetweenPrisoners` still returns it and the migrated NA is correctly closed rather than left open.

The two existing tests (`can merge prisoner numbers` and `can move a booking between prisoner numbers`) were updated to mock `findOpenBetweenPrisoners` instead of the old `findByFirstPrisonerNumberAndSecondPrisonerNumber`.

---

## Files changed

| File | Change |
|------|--------|
| `src/main/kotlin/.../jpa/repository/NonAssociationsRepository.kt` | Added `findOpenBetweenPrisoners` JPQL query |
| `src/main/kotlin/.../service/NonAssociationsMergeService.kt` | Replaced directional single-result lookup with bidirectional open-only lookup |
| `src/test/kotlin/.../service/NonAssociationsMergeServiceTest.kt` | Updated mocks; added two regression tests |

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/NonAssociationsRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/jpa/repository/NonAssociationsRepository.kt
@@ -52,12 +52,24 @@ interface NonAssociationsRepository : JpaRepository<NonAssociation, Long> {
     @Param("isClosed") isClosed: Boolean,
   ): List<NonAssociation>
 
-  // TODO: Assumes that there can only be 1 non-association given a pair of prisoner numbers.
-  //       In future, this will only be true for open non-associations.
-  fun findByFirstPrisonerNumberAndSecondPrisonerNumber(
-    firstPrisonerNumber: String,
-    secondPrisonerNumber: String,
-  ): NonAssociation?
+  /**
+   * Finds all open non-associations between two prisoners, regardless of which is first or second.
+   * Used during merge to detect duplicates without being tripped up by closed historical records
+   * or direction-of-storage differences.
+   */
+  @Query(
+    """
+      SELECT n FROM NonAssociation n
+      WHERE n.isClosed = false AND (
+        (n.firstPrisonerNumber = :firstPrisonerNumber AND n.secondPrisonerNumber = :secondPrisonerNumber) OR
+        (n.firstPrisonerNumber = :secondPrisonerNumber AND n.secondPrisonerNumber = :firstPrisonerNumber)
+      )
+    """,
+  )
+  fun findOpenBetweenPrisoners(
+    @Param("firstPrisonerNumber") firstPrisonerNumber: String,
+    @Param("secondPrisonerNumber") secondPrisonerNumber: String,
+  ): List<NonAssociation>
 }
 
 /**
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsnonassociationsapi/service/NonAssociationsMergeService.kt
@@ -70,36 +70,34 @@ class NonAssociationsMergeService(
       .forEach { nonAssociation ->
         log.debug("Looking at record {}", nonAssociation)
         if (nonAssociation.firstPrisonerNumber == oldPrisonerNumber) {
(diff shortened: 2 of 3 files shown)


steverendell (approved):
LGTM!

xoen (approved):

